PR #3, opened 2016-11-04
Adding the necessary infrastructure for writing unit tests

darthvader88:
This change improves build system by moving the build of each sub-product into a separate subdir in build/. Also, it adds a target for building unit tests that build its own 'test_kernel' with different flags.
The idea is that unit tests will use a special "test" build of the kernel for testing it.
For the moment, no tests are present.

As a side-effect of this change, the -fleading-underscore has been removed from the CFLAGS, making the code nicer. This was necessary in order to the kernel objects to be linked with the regular executable 'unittests'.


src/kernelAsm.asm
```diff
@@ -1,13 +1,13 @@
...
 
-extern _kmain
-extern _idtp
-extern _irq_handler
-extern _gdt_pointer
-
-global _gdt_load
-global _idt_load
-global _tss_flush
+extern kmain
```
ZombineDev: The code indeed lookz nicer without underscores.

unittests/asmMocks.c
```diff
@@ -0,0 +1,55 @@
+void isr0() { }
```
ZombineDev: Yeah I had something like that in mind when we talked about making the code more testable.

unittests/testmain.c
```diff
@@ -0,0 +1,11 @@
+#include <stdio.h>
+
+void *kmalloc(size_t);
+
+int main(int argc, char **argv) {
+
+   printf("hello from the kernel unit tests!\n");
+   printf("kmalloc(10) returns: %p\n", kmalloc(10));
+
```
ZombineDev: void *ptr = kmalloc(10);
assert (ptr != 0);

src/kernelAsm.asm
darthvader88: Yep!

unittests/testmain.c
darthvader88: That was not a test. That's why I did not use any assert.
My point was to try calling an independent kernel function from the tests and see that it works.

PR #29, opened 2017-05-06
Sync primitives

darthvader88:


ZombineDev:
Your making great progress, Vlad. Thanks for keeping me updated by using pull requests!

darthvader88:
Thanks to you for all the nice ideas like using a CI system for Github! I'm so happy with Travis!

PR #47, opened 2018-01-06
Working on the file system support

vvaltchev:


ZombineDev:
Great stuff, Vlad! Keep it going :P

vvaltchev:
> Great stuff, Vlad! Keep it going :P

Thanks so much!!
I'm happy that you are still following my project :-)

vvaltchev:
Actually, if you look at the final version of the code, the loop become larger:

   for (i = 0; i < ARRAY_SIZE(mps); i++) {

      if (!mps[i])
         break;

      if (mps[i]->fs == fs) {
         return -1;
      }

      if (!strcmp(mps[i]->path, path)) {
         return -2;
      }
   }

PR #58, opened 2019-06-17
[scripts]: Make the bash scripts a bit more portable

ZombineDev:
Use `#!/usr/bin/env bash`, instead of `#!/bin/bash` as bash is not in the `/bin/` dir in all *nix systems.


vvaltchev:
Thanks Petar!

PR #66, opened 2020-11-06
Added Lua support

Narasimha1997:
Hey!
 I got interested in this project and I would like to make a small contribution.
I am making this PR to add Lua interpreter support. I have added all the necessary scripts in `scripts/bash_includes/tc/lua` and also I have added `add_lua` routine in `templates/build_fatpart` .

As you know Lua is a tiny interpreter and it is a cool nice feature to have in the toolchain. Having Lua in the toolchain would allow anyone to build cool new features with Lua scripting language. (Most of the hobby operating systems I have seen has a Lua port).

None of the kernel functions are modified as a part of this PR.

I have tried my best to maintain the same coding standards as yours, so that there will be a uniformity.
Please review the PR, I am happy to make any changes you request.

Regards,
Prasanna

### DIFF
--- a/other/toolchain_builds.yml
+++ b/other/toolchain_builds.yml
@@ -49,6 +49,8 @@ steps:
     displayName: "[toolchain] Build vim"
   - script: sudo -E ./scripts/build_toolchain -s build_tfblib
     displayName: "[toolchain] Build tfblib"
+  - script: sudo -E ./scripts/build_toolchain -s build_lua
+    displayName: "[toolchain] Build Lua"
   - script: make -j
     displayName: Build the kernel
   - script: make -j gtests
--- a/scripts/bash_includes/tc/lua
+++ b/scripts/bash_includes/tc/lua
@@ -1,0 +1,59 @@
+#!/usr/bin/env bash
+# SPDX-License-Identifier: BSD-2-Clause
+
+function download_lua {
+   local version="5.4.1"
+   local filename="lua"
+   local tarname="lua-$version.tar.gz"
+   local url="https://www.lua.org/ftp/"
+
+   download_file_in_cache "$url" "$tarname" lua
+   extract_cachefile_tar_gz "$tarname" "lua-$version" "lua"
+}
+
+all_funcs_list+=(build_lua)
+function build_lua {
+
+   pushd $ARCH
+   reset_cc_vars
+
+   show_work_on_component_msg "LUA"
+
+   if ! [ -d "lua" ]; then
+
+      download_lua
+      cd lua
+
+      if [[ "$HOST_ARCH" != "$BUILD_ARCH" ]]; then
+
+         # Case HOST_ARCH == ARCH
+         # Note: if HOST_ARCH not in [BUILD_ARCH, ARCH], we fail earlier.
+         #
+         # Note: NOT TESTED yet.
+         # Remember: this scenario (building on $ARCH) is _not_ officially
+         # supported.
+
+         export SYS_CC="$CC"
+         export SYS_CXX="$CXX"
+         export CC_POST_FLAGS="-specs $MUSL_INSTALL/lib/musl-gcc.specs"
+      else
+         set_cc_vars_to_tc
+      fi
+
+      echo "Building lua..."
+      make clean
+      #build lua and luac static binaries
+      local flags=""
+      flags="$flags CC=\"$CC\" MYCFLAGS=\"-std=gnu99\""
+      flags="$flags AR=\"$AR rcu\" RANLIB=\"$RANLIB\""
+      flags="$flags MYLDFLAGS=\"-static\""
+      run_command2 "make $flags" build.log
+
+      echo "Lua built successfully"
+
+   else
+      show_skip_component_msg "LUA"
+   fi
+
+   popd
+}
--- a/scripts/templates/build_fatpart
+++ b/scripts/templates/build_fatpart
@@ -257,6 +257,21 @@ function add_tree_cmd {
    link_or_copy $tc/$arch/tree_cmd/tree usr/bin
 }
 
+function add_lua {
+
+   if [ -z "@EXTRA_LUA_ENABLED@" ]; then
+      return
+   fi
+
+   if [ -f $tc/$arch/lua/src/lua ]; then
+      link_or_copy $tc/$arch/lua/src/lua bin/
+   fi
+
+   if [ -f $tc/$arch/lua/src/luac ]; then
+      link_or_copy $tc/$arch/lua/src/luac bin/
+   fi
+}
+
 function make_fatpart {
 
    create_fatpart_if_necessary
@@ -312,5 +327,6 @@ add_micropython
 add_fbdoom
 add_vim
 add_tree_cmd
+add_lua
 
 make_fatpart
--- a/other/cmake/config_fatpart
+++ b/other/cmake/config_fatpart
@@ -11,3 +11,4 @@ extra = @USERAPPS_extra@
 fbtest = @USERAPPS_fbtest@
 ncapp = @USERAPPS_ncapp@
 termtest = @USERAPPS_termtest@
+lua = @EXTRA_LUA@
--- a/other/cmake/extra_apps.cmake
+++ b/other/cmake/extra_apps.cmake
@@ -26,6 +26,10 @@ if (EXISTS ${TCROOT}/${ARCH}/tree_cmd)
    message(STATUS "EXTRA_TREE_CMD: ${EXTRA_TREE_CMD}")
 endif()
 
+if (EXISTS ${TCROOT}/${ARCH}/lua)
+   set(EXTRA_LUA OFF CACHE BOOL "Load LUA in Tilck")
+   message(STATUS "EXTRA_LUA: ${EXTRA_LUA}")
+endif()
 
 if (EXTRA_VIM)
    set(EXTRA_VIM_ENABLED "1")
@@ -50,4 +54,8 @@ endif()
 
 if (EXTRA_TREE_CMD)
    set(EXTRA_TREE_CMD_ENABLED "1")
+endif()
+
+if (EXTRA_LUA)
+   set(EXTRA_LUA_ENABLED "1")
 endif()


vvaltchev:
Hi Prasanna,
I'm so happy that you took interest in Tilck and that you'd like to contribute.
I totally agree that adding support for Lua is a cool feature.

Said that, as I hope you understand, I have several comments about the patch.
It's all about keeping the same coding style and conventions etc. I'll add those comments
directly near the code, except for the first one: trailing whitespace at the end of several lines
of code. Could you please remove that? My personal advice is just configure your editor to do that
for you. I use Visual Studio Code and Tilck have a .vscode/settings.json to help with that. But yeah,
in general most of the code editors have some option for that.

Thanks a lot,
Vlad

scripts/bash_includes/tc/lua
```diff
@@ -0,0 +1,49 @@
+#!/usr/bin/env bash
+# SPDX-License-Identifier: BSD-2-Clause
+
+function download_lua {
```
vvaltchev, line 4: I'm happy to see that, overall, your patch is aligned with Tilck's code style :+1: 
Of course, there are some details to fix.

1. While there's nothing really wrong with this function, I feel like an overkill to have params. Just hard-code the version in a local variable. Also, no need for a lua="lua" variable.

2. While there's nothing wrong with the "${varname}" syntax, it's just too heavy and unnecessary most of the time. In Tilck, I use naked variables for plain things, like version, tarname etc., while I use the "$var" syntax for paths and URLs.

For example:
```
   download_file_in_cache "$url" $tarname lua
   extract_cachefile_tar_gz $tarname lua-${version} lua
```

scripts/bash_includes/tc/lua
```diff
@@ -0,0 +1,49 @@
...
+   download_file_in_cache "${url}" "${tarname}" lua
+   extract_cachefile_tar_gz "${tarname}" "${lua}-${version}" "${lua}"
+}
+
+all_funcs_list+=(build_lua)
+function build_lua {
+
+   pushd ${ARCH} 
+
+   lua="lua"
```
vvaltchev, line 20: No need for `lua` and `lua_version` variables. See the comment above.

scripts/bash_includes/tc/lua
```diff
@@ -0,0 +1,49 @@
...
+   lua_version="5.4.1"
+   reset_cc_vars
+
+   show_work_on_component_msg "LUA"
+
+   if ! [ -d "${lua}" ]; then
```
vvaltchev, line 26: The idea across all the build toolchain scripts is that they can be run over and over again, without effect except for the first time. They don't support rebuild (except when we manually delete the $ARCH/<app> directory. So, the skeleton for the build_XXX funcs is something like:

```
all_funcs_list+=(build_XXX)
function build_XXX {

   pushd ${ARCH}

   if ! [ -d XXX ]; then

      show_work_on_component_msg "XXX"
      reset_cc_vars

      download_XXX
      cd XXX

      if [[ "$HOST_ARCH" != "$BUILD_ARCH" ]]; then

         # Case HOST_ARCH == ARCH
         # Note: if HOST_ARCH not in [BUILD_ARCH, ARCH], we fail earlier.
         #
         # Note: NOT TESTED yet.
         # Remember: this scenario (building on $ARCH) is _not_ officially
         # supported.

         export SYS_CC="$CC"
         export SYS_CXX="$CXX"
         export CC_POST_FLAGS="-specs $MUSL_INSTALL/lib/musl-gcc.specs"

      else
         # Default case
         set_cc_vars_to_tc
      fi

      #
      # INSERT APP SPECIFIC CONFIG FLAGS HERE
      #
      run_command2 "make $flags" build.log

   else

      show_skip_component_msg "XXX"
   fi

   popd
}
```

scripts/bash_includes/tc/lua
```diff
@@ -0,0 +1,49 @@
...
+
+   cd ${lua}
+
+   set_cc_vars_to_tc
+   echo "Bulding lua..."
+
+   make clean
+   
+   #build lua and luac static binaries
+   make generic CC="${CC} -std=gnu99" \
```
vvaltchev, line 39: Because `run_command2` is used everywhere, this should be turned into something like:
```
      local flags=""
      flags="$flags CC=\"$CC\" MYCFLAGS=\"-std=gnu99\""
      flags="$flags AR=\"$AR rcu\" RANLIB=\"$RANLIB\""
      flags="$flags MYLDFLAGS=\"-static\""
      run_command2 "make $flags" build.log
```

Narasimha1997:
@vvaltchev , Hey thanks for the response. 
 Yes, maintaining the code quality is an important aspect. Thanks. I used sublime, so `.vscode/settings.json` was ignored.
I the latest commit, I have removed trailing whitespaces as you mentioned.

other/toolchain_builds.yml
```diff
@@ -49,6 +49,8 @@ steps:
     displayName: "[toolchain] Build vim"
   - script: sudo -E ./scripts/build_toolchain -s build_tfblib
     displayName: "[toolchain] Build tfblib"
+  - script: sudo -E ./scripts/build_toolchain -s build_lua
```
vvaltchev, line 52: That's perfect! :+1:

vvaltchev (changes requested):

scripts/templates/build_fatpart
```diff
@@ -257,6 +257,16 @@ function add_tree_cmd {
    link_or_copy $tc/$arch/tree_cmd/tree usr/bin
 }
 
+function add_lua {
```
vvaltchev, line 260: I don't know what happened but my long comment here disappeared. Maybe I didn't save it?
So, I'll re-write it, from scratch :-(

With the `add_lua` function written like that, `lua` will be always built in Tilck's image if it's present in the toolchain. At the beginning I used this approach too, but then the amount of apps to be loaded became too big and it's not convenient to remove/re-add an app from the toolchain every time. Therefore, I added a CMake variable for each package. See `add_vim`. In other words, here we need something like:

```
   if [ -z "@EXTRA_LUA_ENABLED@" ]; then
      return
   fi
```

Of course, we have to define that variable as well.
The place for that is `other/cmake/extra_apps.cmake`.
Just add:
```
if (EXISTS ${TCROOT}/${ARCH}/lua)
   set(EXTRA_LUA OFF CACHE BOOL "Load LUA in Tilck")
   message(STATUS "EXTRA_LUA: ${EXTRA_LUA}")
endif()
```
Below `if (EXISTS ${TCROOT}/${ARCH}/tree_cmd)` and then, at the bottom of the file,
add:

```
if (EXTRA_LUA)
   set(EXTRA_LUA_ENABLED "1")
endif()
```

Finally, we have to add an entry like:
```
lua = @EXTRA_LUA@
```

In `other/cmake/config_fatpart` in order to trigger the rebuild of the `fatpart` when the `EXTRA_LUA` option is changed.

Note: the most convenient way I have to change on-the-fly the cached CMake options is to run:
```
ccmake ./build
```

`ccmake` is a simple text-based UI for reconfiguring CMake projects.
Of course, you can use ./scripts/cmake_run -DEXTRA_LUA=1 but with `ccmake` you see all the configuration options at once.

I believe that's all :-)

vvaltchev (changes requested):

other/toolchain_builds.yml
Narasimha1997: Thanks :+1:

scripts/bash_includes/tc/lua
Narasimha1997: Hey, I have updated the function as per the comment.

scripts/bash_includes/tc/lua
Narasimha1997: I have changed the function structure as per the standard skeleton.

scripts/templates/build_fatpart
Narasimha1997: Hey @vvaltchev 
I have modified `other/cmake/extra_apps.cmake`, `other/cmake/config_fatpart` and `scripts/templates/build_fatpart` to include the changes you mentioned. Also I have tested these changes.

Narasimha1997:
@vvaltchev 
I think all the changes you requested has been made. 
I really appreciate the way you are maintaining the code standards and conventions. This is a great learning for me.

Thanks and Regards,
Prasanna

scripts/bash_includes/tc/lua
```diff
@@ -2,48 +2,58 @@
...
+      fi
 
-   #build lua and luac static binaries
-   make generic CC="${CC} -std=gnu99" \
-      MYCFLAGS="-static" \
-      AR="${AR} rcu" RANLIB=${RANLIB} \
-      MYLDFLAGS="-static" \
-      SYSLIBS="-Wl,-E,-strip-all" \
-      &> build.log
+      echo "Bulding lua..."
```
vvaltchev, line 43:
```suggestion
      echo "Building lua..."
```

scripts/bash_includes/tc/lua
```diff
@@ -2,48 +2,58 @@
...
+      make clean
+      #build lua and luac static binaries
+      local flags=""
+      flags="$flags CC=\"$CC\" MYCFLAGS=\"-std=gnu99\""
+      flags="$flags AR=\"$AR rcu\" RANLIB=\"$RANLIB\""
+      flags="$flags MYLDFLAGS=\"-static\""
+      run_command2 "make $flags" build.log
 
-   echo "Built lua successfully"
+      echo "Built lua successfully"
```
vvaltchev, line 52:
```suggestion
      echo "Lua built successfully"
```

vvaltchev (changes requested):
Just one last thing: fix two echo messages.
Also, can you squash all the commits into one?
Note: for commits like that, I the following format for the commit message:

"[toolchain] Add the XXX extra package"

vvaltchev:
> @vvaltchev
> I think all the changes you requested has been made.
> I really appreciate the way you are maintaining the code standards and conventions. This is a great learning for me.
> 
> Thanks and Regards,
> Prasanna

@Narasimha1997 
And I really appreciate your approach towards contributing to this project.
Thank you so much for that. If you're interested in contributing more, just let
me know: you can both work of whatever you like in Tilck or, if you need
ideas, I have a list of TODOs as well :-)

Vlad

P.S. I made a last review with just minor fixes, after which I'll be happy to immediately
merge your patch.

scripts/bash_includes/tc/lua
Narasimha1997: @vvaltchev 
Fixed these typos and squashed everything into a single commit. Please checkout.

Narasimha1997:
@vvaltchev 
Hey, sure.
It would be helpful if you share the roadmap.

Thanks and Regards,
Prasanna

PR #73, opened 2021-07-06
[scripts] Fix WSL detection via uname

ytret:
I am using Windows Subsystem for Linux (version 2), and several scripts in `build/` could not detect that to warn me that I cannot use KVM. The reason is that in my case the output of `uname -a` contains `microsoft` (completely lowercase) rather than `Microsoft`:

```
$ uname -a
Linux DESKTOP-78BKHO2 5.4.72-microsoft-standard-WSL2 #1 SMP Wed Oct 28 23:40:43 UTC 2020 x86_64 x86_64 x86_64 GNU/Linux
```

I have changed the scripts so that they do a case-independent search for "microsoft", and now they prevent me from running KVM scripts.

### DIFF
--- a/scripts/templates/qemu/run_efi_qemu32
+++ b/scripts/templates/qemu/run_efi_qemu32
@@ -2,7 +2,7 @@
 
 BUILD_DIR="@CMAKE_BINARY_DIR@"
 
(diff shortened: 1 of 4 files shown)


vvaltchev:
Thanks for finding and fixing this issue!

PR #106, opened 2022-07-23
[cmake-configurator] First steps: introduce the parser

michelececcacci:
Business logic for cmake-configurator, it's unit tested and has  unit test/coverage  support for CI/CD.

### DIFF
--- a/other/parser_build.yml
+++ b/other/parser_build.yml
@@ -1,0 +1,35 @@
+trigger: none
+pr: parser
+
+pool:
+  vmImage: 'ubuntu-18.04'
+strategy:
+  matrix:
+    Python36:
+      python.version: '3.6'
+    Python37:
+      python.version: '3.7'
+    Python38:
+      python.version: '3.8'
+    Python39:
+      python.version: '3.9'
+    Python310:
+      python.version: '3.10'
+
+steps:
+  - task: UsePythonVersion@@
+    inputs:
+      versionSpec: '$(python.version)'
+    displayName: 'Use Python $(python.version)'
+
+  - script:
+      cd scripts/parser/backend |
+      pip3 install coverage
+      displayName: setup
+  - script:
+      coverage run -m unittest discover |
+    displayName: run unit tests on parser
+    script:
+      coverage report |
+      bash <  (curl -s https://codecov.io/bash)
+    displayName: generate report coverage report and send to codecov
--- a/scripts/cmake_configurator/parser/cmake_parser.py
+++ b/scripts/cmake_configurator/parser/cmake_parser.py
@@ -1,0 +1,38 @@
+from cmake_variable import cmake_variable
+from cmake_string import cmake_string
+
+class cmake_parser:
+   def parse(self, file_path: str) -> list[cmake_variable]:
+      """
+      this will communicate with the gui business logic, and pass to it
+      variables ready to be displayed
+      """
+      comment: str = ""
+      all_vars: list[cmake_variable] = []
+      self.file_lines = []
+      with open(file_path) as f:
+         self.file_lines = f.readlines()
+
+      for i, line in enumerate(self.file_lines):
+         parsed_string: cmake_string = cmake_string(line)
+         if parsed_string.is_variable():
+            j = i
+            while j >=0 and cmake_string(self.file_lines[j]).is_comment():
+               comment = cmake_string(self.file_lines[j]).user_comment() +  comment
+               j -= 1
+            all_vars.append(cmake_variable(line, comment, i))
+
+      return all_vars
+
+   def write(self, new_vars: list[cmake_variable], file_path: str) -> None:
+      """
+      this will take the modified input from the gui and write it to file, then
+      the gui will re-read  and display the file contents in order to have an
+      updated version. Since we assume new data will come from the GUI,
+      here we assume that the file is already parsed.
+      """
+      for var in new_vars:
+         self.file_lines[var.row_number] = var.serialize_line()
+
+      with open(file_path, "w+") as f:
+         f.writelines(self.file_lines)
--- a/scripts/cmake_configurator/parser/cmake_string.py
+++ b/scripts/cmake_configurator/parser/cmake_string.py
@@ -1,0 +1,41 @@
+import re
+
+class cmake_string(str):
+   """
+   simple utility string that abstracts away some details from the parsing
+   process
+   """
+   regex_expression =  r"^([A-Za-z_0-9]+?):([A-Za-z]+?)=(.*)$"
+
+   def user_comment(self) -> str:
+      """
+      if a line starts with // we assume it's a comment, and we strip away the
+      newline that is added by CMake
+      """
+      return self[2:].rstrip()
+
+   def is_comment(self) -> bool:
+      return self[:2] == "//" or self[0] == "#"
+
+   def is_empty(self) -> bool:
+      """
+      lenght assumption is safe in this case because of the CMakeCache format
+      A comment has at least lenght 2, while a variable has always a name and
+      an explicit type
+      """
+      return len(self) <= 1
+
+   def parse_variable(self):
+      """
+      match has three groups, the first is variable name, the second is its
+      type and the third is its value
+      """
+
+      # rstrip is needed for 3rd group
+      m = re.match(self.regex_expression, self.rstrip())
+      return m
+
+   def is_variable(self) -> bool:
+
+      m = self.parse_variable()
+      return (m and len(m.groups()) >= 3)
--- a/scripts/cmake_configurator/parser/cmake_variable.py
+++ b/scripts/cmake_configurator/parser/cmake_variable.py
@@ -1,0 +1,37 @@
+"""
+basic idea here is just having a special variable that deals with i/o
+and validation, while this just communicates with TUI.
+"""
+from cmake_string import cmake_string
+
+
+class cmake_variable:
+   BOOL_MAPPINGS = {
+      "ON" : True,
+      "OFF": False,
+      "0" : True,
+      "1" : False,
+      "TRUE": True,
+      "FALSE": False,
+   }
+
+   def __init__(self, line: str, comment: str, row_number: int) -> None:
+      self.comment: str = comment
+      self.row_number: row_number = row_number
+      s = cmake_string(line)
+      self.name, self.type, self.val =  s.parse_variable().groups()
+
+   def is_valid(self, user_input: str) -> bool:
+      """
+      TODO: this might need to be extened in the future.
+      """
+      if self.type == "BOOL":
+         return user_input.upper() in self.BOOL_MAPPINGS
+      elif self.type == "INTERNAL":
+         return False
+      return True
+
+
+   def serialize_line(self) -> str:
+      "returns string in the correct format to be dumped to file"
+      return self.name + ":" + self.type + "=" + self.val
--- a/scripts/cmake_configurator/parser/test_cmakestring.py
+++ b/scripts/cmake_configurator/parser/test_cmakestring.py
@@ -1,0 +1,45 @@
+import unittest
+from cmake_string import cmake_string
+
+class test_cmake_string(unittest.TestCase):
+   def __init__(self, methodName: str = ...) -> None:
+      super().__init__(methodName)
+      self.slash_comment: str  = "//Build unit tests for the target arch"
+      self.not_comment: str = "Build unit tests for the target arch"
+      self.hashtag_comment: str = "#Build unit tests for the target arch"
+
+   def test_comment_string(self):
+      string = cmake_string(self.slash_comment)
+      self.assertEqual(string.user_comment(), self.slash_comment[2:])
+
+   def test_is_comment_slash(self):
+      string = cmake_string(self.slash_comment)
+      self.assertTrue(string.is_comment())
+
+   def test_is_comment_hashtag(self):
+      string = cmake_string(self.hashtag_comment)
+      self.assertTrue(string.is_comment())
+
+   def test_is_not_comment(self):
+      string = cmake_string(self.not_comment)
+      self.assertFalse(string.is_comment())
+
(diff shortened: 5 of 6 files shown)


scripts/parser/backend/cmakecachestring.py
```diff
@@ -0,0 +1,41 @@
...
+
+   def user_comment(self) -> str:
+      """
+      if a line starts with // we assume it's a comment, and we strip away the
+      newline that is added by CMake
+      """
+      return self[2:].rstrip()
+
+   def is_comment(self) -> bool:
+      return self[:2] == "//"
```
vvaltchev, line 18: OK, but what about the comments starting with '#'? How do you skip them?

scripts/parser/backend/cmakecachestring.py
```diff
@@ -0,0 +1,41 @@
...
+
+   def is_empty(self) -> bool:
+      """
+      lenght assumption is safe in this case because of the CMakeCache format
+      A comment has at least lenght 2, while a variable has always a name and
+      an explicit type
+      """
+      return len(self) <= 1
+
+   def split_variable(self):
```
vvaltchev, line 28: **Misleading name**: This should really be something like `parse_variable`.

scripts/parser/backend/cmakecachevariable.py
```diff
@@ -0,0 +1,39 @@
...
+      s = cmc_str(line)
+      self.name, self.type, self.val =  s.split_variable().groups()
+
+   def is_valid(self, user_input: str) -> bool:
+      """
+      TODO: this might need to be extened in the future.
+      """
+      if self.type == "BOOL":
+         return user_input in self.BOOL_MAPPINGS
+      elif self.type == "INT":
```
vvaltchev, line 30: CMake doesn't have an INT type. It's either BOOL, STRING, FILEPATH or INTERNAL.

scripts/parser/backend/cmakecachevariable.py
```diff
@@ -0,0 +1,39 @@
...
+      self.idx: idx = idx
+      s = cmc_str(line)
+      self.name, self.type, self.val =  s.split_variable().groups()
+
+   def is_valid(self, user_input: str) -> bool:
+      """
+      TODO: this might need to be extened in the future.
+      """
+      if self.type == "BOOL":
+         return user_input in self.BOOL_MAPPINGS
```
vvaltchev, line 29: Either here on in BOOL_MAPPINGS, you need to consider the lower case versions as well. So, include `on`, `off`, `true`, `false`.

other/parser_build.yml
```diff
@@ -0,0 +1,35 @@
+trigger: none
```
vvaltchev, line 1: You should add here an expression path to be triggered automatically when files in the cmake-configurator are modified.
Also, how did you test this YAML file? Did you create a dedicated pipeline in your Azure account?

other/parser_build.yml
```diff
@@ -0,0 +1,35 @@
...
+pr: parser
```
vvaltchev, line 2: What does this do?

other/parser_build.yml
vvaltchev, line 1: **Misleading name**: Also, don't like the name "parser" here. Tilck is medium-sized project with a ton of stuff inside. Parser doesn't mean anything. The file name should be something like `cmake_configurator.yml`

scripts/parser/backend/cmakecacheparser.py
```diff
@@ -0,0 +1,37 @@
+from os.path import dirname
```
vvaltchev, line 1: As before, everything here is in `script/parser` which means nothing in the context of Tilck. you should create a directory like `scripts/cmake_configurator` and put there a `parser` sub-directory, instead of using `backend`.

scripts/parser/backend/cmakecacheparser.py
```diff
@@ -0,0 +1,37 @@
...
+      """
+      this will take the modified input from the gui and write it to file, then
+      the gui will re-read  and display the file contents in order to have an
+      updated version. Since we assume new data will come from the GUI,
+      here we assume that the file is already parsed.
+      """
+      for var in new_vars:
+         self.file_lines[var.idx] = var.serialize_line()
+
+      with open(file_path, "w+") as f:
```
vvaltchev, line 36: If `file_lines` becomes an array of cmakecache_row objects, you could have a method serialize() there, which will call var.serialize() in case it's a variable.

scripts/parser/backend/cmakecachestring.py
```diff
@@ -0,0 +1,41 @@
...
+      """
+      match has three groups, the first is variable name, the second is its
+      type and the third is its value
+      """
+
+      # rstrip is needed for 3rd group
+      m = re.match(self.regex_expression, self.rstrip())
+      return m
```
vvaltchev, line 36: Here it would really make sense to return a `cmakecache_variable` object or None, not the result of the regex match.

scripts/parser/backend/cmakecachevariable.py
```diff
@@ -0,0 +1,39 @@
...
+      "OFF": False,
+      "0" : True,
+      "1" : False,
+      "TRUE": True,
+      "FALSE": False,
+   }
+
+   def __init__(self, line: str, comment: str, idx: int) -> None:
+      self.comment: str = comment
+      self.idx: idx = idx
```
vvaltchev, line 20: **Misleading name**: Index does not mean anything to me. It really should be something like `row_number` or just `row`.

scripts/parser/backend/test_cmakecachevariable.py
```diff
@@ -0,0 +1,12 @@
+import unittest
+from cmakecachevariable import cmakecache_variable as cmc_var
```
vvaltchev, line 2: Even if we're talking all the time about cmakecache variables, putting this "cache" suffix _everywhere_ creates a lot of clutter. Isn't better to just use cmake_var? This will save us from using cmc_var etc.

scripts/parser/backend/test_cmakecachevariable.py
```diff
@@ -0,0 +1,12 @@
...
+
+class test_cmake_cache_variable(unittest.TestCase):
```
vvaltchev, line 4: That's OK, but one of the things that really needed to be tested here is the parser itself. In particular, the part which extracts all the comments above the variables. I'd like to see a test where, given multi-line strings, they are parsed and the result is checked. For example:

1. case without comments, does it work?
```
A:STRING=val
B:STRING=val2
```
2. case with comments and spaces
```
// this is a comment for variable A
// second line

A:STRING=val

// this is a comment for variable B
B:STRING=val2
```

To achieve this, you will need to refactor the parser class and extract the method which concatenates all the comments above a variable.

vvaltchev (changes requested):
Thanks for working on this, Michele!

scripts/parser/backend/cmakecachevariable.py
michelececcacci: I changed `idx` to `row_number`.

scripts/parser/backend/cmakecachevariable.py
michelececcacci: Eliminated the INT type.

scripts/parser/backend/test_cmakecachevariable.py
michelececcacci: Done.

scripts/parser/backend/cmakecachestring.py
michelececcacci: Currently using, 
 ```python
  return self[:2] == "//" or self[0] == "#"
```
should i add a unit test for it?

File: scripts/parser/backend/cmakecacheparser.py [hunk 13]
```diff
@@ -0,0 +1,37 @@
+from os.path import dirname
+from cmakecachevariable import cmakecache_variable as cmc_var
+from cmakecachestring import cmakecache_string as cmc_str
+
+class cmake_cache_parser:
```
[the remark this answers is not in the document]
michelececcacci: Resolved. Using all underscores now.

scripts/parser/backend/cmakecacheparser.py
```diff
@@ -0,0 +1,37 @@
...
+   def parse(self, file_path: str) -> list[cmc_var]:
+      """
+      this will communicate with the gui business logic, and pass to it
+      variables ready to be displayed
+      """
+      comment: str = ""
+      self.file_lines = open(dirname(__file__) + file_path).readlines()
```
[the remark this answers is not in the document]
michelececcacci: Resolved. Now using just the `file_path` variable

scripts/parser/backend/cmakecachestring.py
michelececcacci: changed.

scripts/parser/backend/cmakecacheparser.py
michelececcacci: Done.

scripts/parser/backend/cmakecachevariable.py
michelececcacci: Should i use `self.upper` to avoid code duplication or add the lowercase types in the mappings?

scripts/parser/backend/cmakecachestring.py
vvaltchev: Sure, why not.

scripts/parser/backend/cmakecachevariable.py
vvaltchev: yes, it's better to flatten the case with either upper() or lower() on both sides.

scripts/cmake_configurator/parser/cmake_parser.py
```diff
@@ -9,7 +8,7 @@ def parse(self, file_path: str) -> list[cmake_variable]:
       variables ready to be displayed
       """
       comment: str = ""
-      self.file_lines = open(dirname(__file__) + file_path).readlines()
+      self.file_lines = open(file_path).readlines()
```
vvaltchev, line 11: Please squash this commit into the other one. Also, what happens with the file handle returned by open(..) ? When do we close it? Shouldn't we use with open(...) as fd etc. ?

vvaltchev (changes requested):

scripts/cmake_configurator/parser/cmake_parser.py
michelececcacci: You are right on both of those. I'm fixing it right now.

scripts/parser/backend/cmakecachestring.py
michelececcacci: Done.

scripts/cmake_configurator/parser/cmake_string.py
```diff
@@ -0,0 +1,41 @@
+import re
+
+class cmake_string(str):
```
vvaltchev, line 3: OK, let me explain here the idea, other than using `str` as data and not inheriting from it and calling the class `cmake_row` (or something like it): the row should be parsed in the construtor `__init__(self, row)`. So there shouldn't exist things like parse_variable(). You parse the row ONCE and then you can get its type (comment, empty, variable, other) and it's content (variable object, text in comment etc.). Get it?

scripts/cmake_configurator/parser/cmake_string.py
```diff
@@ -0,0 +1,41 @@
...
+      """
+      if a line starts with // we assume it's a comment, and we strip away the
+      newline that is added by CMake
+      """
+      return self[2:].rstrip()
+
+   def is_comment(self) -> bool:
+      return self[:2] == "//" or self[0] == "#"
+
+   def is_empty(self) -> bool:
```
vvaltchev, line 20: Please, delete or fix this.

scripts/cmake_configurator/parser/cmake_string.py
```diff
@@ -0,0 +1,41 @@
...
+      """
+      lenght assumption is safe in this case because of the CMakeCache format
+      A comment has at least lenght 2, while a variable has always a name and
+      an explicit type
+      """
+      return len(self) <= 1
+
+   def parse_variable(self):
```
vvaltchev, line 28: Check the comment on the top of the class

vvaltchev (changes requested):

scripts/parser/backend/cmakecachestring.py
```diff
@@ -0,0 +1,41 @@
...
+      """
+      if a line starts with // we assume it's a comment, and we strip away the
+      newline that is added by CMake
+      """
+      return self[2:].rstrip()
+
+   def is_comment(self) -> bool:
+      return self[:2] == "//"
+
+   def is_empty(self) -> bool:
```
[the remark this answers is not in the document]
michelececcacci: Corrected

scripts/cmake_configurator/parser/cmake_string.py
michelececcacci: Corrected

scripts/parser/backend/cmakecacheparser.py
```diff
@@ -0,0 +1,37 @@
...
+
+      all_vars: list[cmc_var] = []
+      for i, line in enumerate(self.file_lines):
+         parsed_string: cmc_str = cmc_str(line)
+         if parsed_string.is_variable():
+            j = i
+            while j >=0 and cmc_str(self.file_lines[j]).is_comment():
```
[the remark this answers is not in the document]
michelececcacci: Noted.

scripts/parser/backend/cmakecachestring.py
michelececcacci: Restructured my code, this is in the parsing of the cmake_row

scripts/cmake_configurator/parser/cmake_string.py
michelececcacci: Done

PR #109, opened 2022-07-27
[cmake-configurator] First steps: introduce the parser

michelececcacci:
Updated #106 with better software design, and also made sure that CI/CD works

### DIFF
--- a/scripts/cmake-configurator/parser/cmake_parser.py
+++ b/scripts/cmake-configurator/parser/cmake_parser.py
@@ -1,0 +1,48 @@
+from cmake_row import cmake_row
+from cmake_variable import cmake_variable
+from typing import List
+
+class cmake_parser:
+   def read_file(self, abs_file_path: str) -> None:
+      "reads file lines and inserts them into file_lines"
+      self.file_lines: List[str] = []
+      with open(abs_file_path) as f:
+         self.file_lines = f.readlines()
+
+   def lines_to_vars(self, file_lines: List[str]) -> List[cmake_variable]:
+      """
+      this will communicate with the gui business logic, and pass to it
+      variables ready to be displayed
+      """
+      self.file_lines: List[str] = file_lines
+      cmake_rows :List[cmake_row] =  []
+      cmake_vars :List[cmake_variable] = []
+
+      for i, line in enumerate(self.file_lines):
+         cmake_rows.append(cmake_row(line, i))
+
+      for i, row in enumerate(cmake_rows):
+         if row.type == cmake_row.VARIABLE:
+            comment: str = ""
+            j = i-1
+            while j >= 0 and cmake_rows[j].type == cmake_row.SLASH_COMMENT:
+               comment = cmake_rows[j].val + comment
+               j -=1
+            cmake_vars.append(cmake_variable(row, comment))
+      return cmake_vars
+
+   def vars_to_lines(self, cmake_vars: List[cmake_variable]) -> List[str]:
+      """
+      this will take the modified input from the gui and write it to file, then
+      the gui will re-read  and display the file contents in order to have an
+      updated version. Since we assume new data will come from the GUI,
+      here we assume that the file is already parsed.
+      """
+      for var in cmake_vars:
+         if var.row.type == cmake_row.VARIABLE:
+            self.file_lines[var.row.row_number] = var.row.serialize()
+      return self.file_lines
+
+   def write_to_file(self, abs_file_path: str) -> None:
+      with open(abs_file_path, "w+") as f:
+         f.writelines(self.file_lines)
--- a/scripts/cmake-configurator/parser/test_cmakeparser.py
+++ b/scripts/cmake-configurator/parser/test_cmakeparser.py
@@ -1,0 +1,66 @@
+from cmake_parser import cmake_parser
+from unittest import TestCase
+
+file_lines_generic = """
+# KEY:TYPE=VALUE
+# KEY is the name of a variable in the cache.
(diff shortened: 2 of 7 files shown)


scripts/cmake-configurator/parser/cmake_parser.py
```diff
@@ -0,0 +1,48 @@
...
+         cmake_rows.append(cmake_row(line, i))
+
+      for i, row in enumerate(cmake_rows):
+         if row.type == cmake_row.VARIABLE:
+            comment: str = ""
+            j = i-1
+            while j >= 0 and cmake_rows[j].type == cmake_row.SLASH_COMMENT:
+               comment = cmake_rows[j].val + comment
+               j -=1
+            cmake_vars.append(cmake_variable(row, comment))
```
vvaltchev, line 31: That's cool, but we're saving the row-comments as they are, without any post-processing.
We should (now, or in the next commit) strip the "//" and append all comment lines into a single line. Keeping the comment as it is now, it doesn't make much sense.

scripts/cmake-configurator/parser/test_cmakeparser.py
```diff
@@ -0,0 +1,45 @@
...
+# KEY:TYPE=VALUE
+# KEY is the name of a variable in the cache.
+# TYPE is a hint to GUIs for the type of VALUE, DO NOT EDIT TYPE!.
+# VALUE is the current value for the KEY.
+
+########################
+# EXTERNAL cache entries
+########################
+
+//Build unit tests for the target arch
```
vvaltchev, line 14: - Missing multi-line comments on the top of variables
- Missing variables without comments on the top

scripts/cmake-configurator/parser/test_cmakerow.py
```diff
@@ -0,0 +1,44 @@
...
+      with self.subTest():
+         self.assertEqual(row.var_type, "FILEPATH")
+      with self.subTest():
+         self.assertEqual(row.type, row.VARIABLE)
+      with self.subTest():
+         self.assertEqual(row.val,  "/usr/bin/addr2line")
+      with self.subTest():
+         self.assertEqual(row.row_number, 0)
+
+   def test_slash_comment_row(self):
```
vvaltchev, line 24: That's cool, but where do we check the contents of the "comment" field? The idea is that in the next step the comments will be very important for us, because they will contain metadata. So, could you compare the value of "comment" with some expected value? Note: don't forget the multi-lib comment case which is essential.

scripts/cmake-configurator/parser/test_cmakevariable.py
```diff
@@ -0,0 +1,28 @@
...
+comment: str = "Build unit tests for the target arch"
+bool_line: str = "ARCH_GTESTS:BOOL=OFF"
+internal_line: str = "CMAKE_AR:INTERNAL=/usr/bin/ar"
+arch_row: cmake_row = cmake_row(bool_line, 0)
+
+class test_cmake_cache_variable(unittest.TestCase):
+
+   def test_var_creation(self):
+      var = cmake_variable(arch_row, comment)
+      self.assertEqual(var.row.name, "ARCH_GTESTS") \
```
vvaltchev, line 14: Why using backslashes and forming a single giant statement? Isn't possible to just write multiple `assertEqual` statements, each on a new line?

scripts/cmake-configurator/parser/test_cmakeparser.py
```diff
@@ -0,0 +1,45 @@
...
+ARCH_GTESTS:BOOL=OFF
+
+//Build the EFI bootloader
+BOOTLOADER_EFI:BOOL=ON
+
+//Build the legacy bootloader
+BOOTLOADER_LEGACY:BOOL=ON
```
vvaltchev, line 21: It would be cooler to be exhaustive at least for the booleans: just create a new variable for each possible bool value. There will be just 6 of them. You can call them BOOLVAR1, BOOLVAR2 etc. Then in the tests, check that all the logical true values map to true and all the logical false values map to false.

scripts/cmake-configurator/parser/test_cmakevariable.py
```diff
@@ -0,0 +1,28 @@
...
+         and self.assertEqual(var.val == "OFF") \
+         and self.assertEqual(var.type == "BOOL")
+
+   def test_input_valid_bool(self):
+      var = cmake_variable(arch_row, comment)
+      with self.subTest():
```
vvaltchev, line 20: Cool, but it would be even cooler to check for all 6 cases. See the comment above.

vvaltchev (changes requested):
Almost completely good, there are just a few minor things to fix.

scripts/cmake-configurator/parser/test_cmakevariable.py
michelececcacci: I used `unittest.subTest()`

scripts/cmake-configurator/parser/cmake_parser.py
michelececcacci: This is already done by `cmake_row`'s constructor

vvaltchev: Ah, OK, that's great! I just didn't notice it.

scripts/cmake-configurator/parser/test_cmakeparser.py
michelececcacci: Resolved

scripts/cmake-configurator/parser/test_cmakerow.py
michelececcacci: Done

PR #113, opened 2022-07-29
[azure] Updated coverage report

michelececcacci:


scripts/templates/generate_coverage_report
```diff
@@ -125,24 +129,38 @@ if [[ "$1" == "--codecov" ]]; then
       echo "WARNING: No CODECOV_TOKEN defined. Upload to CodeCov might fail"
    fi
 
-   # Remove other files to be sure CodeCov uses only the coverage.info file
-   rm -f lcov.info
-   remove_gcda_files
-
    # Upload report to CodeCov
+   if [ ! -d build/tmp ]; then
```
vvaltchev, line 133: Style: `if ! [ -d ....`

scripts/templates/generate_coverage_report
```diff
@@ -112,6 +112,10 @@ if [[ "$1" == "--codecov" ]]; then
       exit 1
    fi
 
+   # Remove other files to be sure CodeCov uses only the coverage.info file
```
vvaltchev, line 115: Unnecessary change: these lines were blow the TOKEN check. Please, don't move them around or the diff gets tricker than necessary to review.

scripts/templates/generate_coverage_report
vvaltchev: **BUG**: Don't mention the "build" directory like that: there is the `$BUILD` variable specifically for that. See above.
The build directory doesn't have to be `TILCK_ROOT/build`

scripts/templates/generate_coverage_report
```diff
@@ -125,24 +129,38 @@ if [[ "$1" == "--codecov" ]]; then
       echo "WARNING: No CODECOV_TOKEN defined. Upload to CodeCov might fail"
    fi
 
-   # Remove other files to be sure CodeCov uses only the coverage.info file
-   rm -f lcov.info
-   remove_gcda_files
-
    # Upload report to CodeCov
+   if [ ! -d build/tmp ]; then
+      mkdir build
```
vvaltchev, line 134: Same as above. Use mkdir -p "$BUILD/tmp"

scripts/templates/generate_coverage_report
```diff
@@ -125,24 +129,38 @@ if [[ "$1" == "--codecov" ]]; then
...
       fi
+
    fi
 
+   chmod +x codecov
+
+   if [ ! -z  $"CODECOV-TOKEN" ]; then
+      ./codecov
+   else
+      ./codecov -t $"CODECOV-TOKEN"
```
vvaltchev, line 162: **BUG:** First, you don't need to do that at all. If the env variable has been exported, the codecov tool will see it.
Also, the name was `CODECOV_TOKEN`. Third, $"something" doesn't mean anything to me. "$something" is the correct form.

scripts/templates/generate_coverage_report
```diff
@@ -125,24 +129,38 @@ if [[ "$1" == "--codecov" ]]; then
...
-   remove_gcda_files
-
    # Upload report to CodeCov
+   if [ ! -d build/tmp ]; then
+      mkdir build
+      mkdir build/tmp
+   fi
 
-   if ! bash <(curl -s https://codecov.io/bash) -X gcov; then
+   cd build/tmp/
```
vvaltchev, line 138: **BUG:** Don't change the directory like that. It might not work.
Do `pushd $BUILD/tmp` and later, after the chmod, do `popd`, to the restore the CWD.
Also, later run codecov from the current directory with: `$BUILD/tmp/codecov`.

vvaltchev (changes requested):

scripts/templates/generate_coverage_report
vvaltchev: **BUG:** the -X gcov option is missing.

michelececcacci: So i just need to do `$BUILD/tmp/codecov -X gcov`?

scripts/templates/generate_coverage_report
michelececcacci: Shouldn't it be "$BUILD"/tmp?

vvaltchev: Absolutely not. Scripting languages like Bash, Perl, PHP and many others (in particular shells), have been designed to expand variables inside strings. In case the next char matches [A-Za-z0-9_], then the correct syntax would be "${BUILD}something". The weird syntax `"$BUILD"/tmp` might work on Bash, but it's generally unportable.

scripts/templates/generate_coverage_report
vvaltchev: Yes, of course. The whole point was to replace `bash <(curl -s https://codecov.io/bash)` with `$BUILD/tmp/codecov` everywhere after downloading it.

scripts/templates/generate_coverage_report
michelececcacci: Ok, thanks for the answer!

scripts/templates/generate_coverage_report
michelececcacci: Perfect!

scripts/templates/generate_coverage_report
```diff
@@ -112,6 +112,7 @@ if [[ "$1" == "--codecov" ]]; then
       exit 1
    fi
 
+
```
vvaltchev, line 115: Extra whitespace, please remove.
Nit: after making a change and saving the commit, just run: `git diff HEAD~1` to inspect your change` and make sure doesn't have any unnecessary things :-)

scripts/templates/generate_coverage_report
```diff
@@ -130,19 +131,35 @@ if [[ "$1" == "--codecov" ]]; then
...
+      curl -Os https://uploader.codecov.io/latest/linux/codecov
+      if ! [ -f codecov ]; then
+         echo "ERROR: codecov did not collect coverage reports"
+         echo "Trying again in 1 minute..."
+         sleep 60
+      fi
 
-      if ! bash <(curl -s https://codecov.io/bash) -X gcov; then
+      curl -Os https://uploader.codecov.io/latest/linux/codecov
```
vvaltchev, line 149: **BUG:** this will run unconditionally, even if the first curl succeeded!
There was a reason why that was under an `if` in the old code.
Please fix it.

scripts/templates/generate_coverage_report
```diff
@@ -130,19 +131,35 @@ if [[ "$1" == "--codecov" ]]; then
...
+      if ! [  -f codecov ]; then
          echo "FATAL ERROR: Codecov cannot collect the coverage reports"
          exit 1
       fi
+
    fi
 
+   chmod +x codecov
+
```
vvaltchev, line 158: style: remove this extra empty line

scripts/templates/generate_coverage_report
```diff
@@ -130,19 +131,35 @@ if [[ "$1" == "--codecov" ]]; then
...
+   popd
+
+   "$BUILD/tmp/codecov" -X gcov
+
```
vvaltchev, line 162: style: remove this extra empty line

scripts/templates/generate_coverage_report
```diff
@@ -130,19 +131,35 @@ if [[ "$1" == "--codecov" ]]; then
...
          echo "FATAL ERROR: Codecov cannot collect the coverage reports"
          exit 1
       fi
+
    fi
 
+   chmod +x codecov
+
+   popd
+
```
vvaltchev, line 160: style: this empty line must be kept, instead.
```
popd
chmod
<empty line>
codecov
exit
```

vvaltchev (changes requested):

vvaltchev (commented):
Great, thanks for fixing!

PR #114, opened 2022-08-03
[cmake-configurator] global refactoring

michelececcacci:
Small improvements to previous cmake-configurator commit, logic is the same but this should be nicer to work with


scripts/cmake-configurator/parser/cmake_row.py
```diff
@@ -32,33 +35,33 @@ def __init__(self, raw_row: str, row_number: int):
       """
       row  = raw_row.strip()
       matches: Optional[Match[str]]  = re.match(self.regex_expression, row)
-      self.type: str =  self.NOT_PARSABLE
+      self.type: str =  types.NOT_PARSABLE.value
```
vvaltchev, line 38: **ISSUE**: What's the point of using an enum if than you're accessing its value? This is not OK. In particular for string values. The `type` field should keep an instance of the Enum class, that's the whole point of having an Enum, right? Also, this way the linter can do checks, otherwise any string would be acceptable.

Also, do you really need to convert the enum type to string somewhere? If you do, it's always possible with a simple dict comprehension expression to generate a value -> enum item string.

scripts/cmake-configurator/parser/cmake_row.py
```diff
@@ -1,5 +1,13 @@
 import re
 from typing import Match, Optional
+from enum import Enum
+
+class types(Enum):
+   EMPTY: str = "EMPTY"
+   SLASH_COMMENT: str = "SLASH_COMMENT"
+   HASHTAG_COMMENT: str = "HASHTAG_COMMENT"
+   VARIABLE: str = "VARIABLE"
+   NOT_PARSABLE: str = "NOT_PARSABLE"
```
vvaltchev, line 10: Using string values for Enum items is **bad style**. See my comments below. Use numbers and keep in `type` an instance of the Enum class, not the string value.

scripts/cmake-configurator/parser/cmake_row.py
```diff
@@ -75,7 +78,13 @@ def serialize(self) -> str:
       returns serialized version of variable. Expects to be called only on
       variable types.
       """
-      return self.name + ":" + self.var_type + "=" + self.val
+      serialized_val = self.val
```
vvaltchev, line 81: **STYLE:** this looks a bit cumbersome. You could re-write it as:
```
if type == "BOOL":
    ser_val = "ON" if self.val else "OFF"
else:
    ser_val = self.val
```

Also, it's really **NOT OK** even if Python scripts to use strings for descripting types. `var_type` should be an Enum as well.

scripts/cmake-configurator/parser/cmake_row.py
```diff
@@ -32,33 +35,33 @@ def __init__(self, raw_row: str, row_number: int):
       """
       row  = raw_row.strip()
       matches: Optional[Match[str]]  = re.match(self.regex_expression, row)
-      self.type: str =  self.NOT_PARSABLE
+      self.type: str =  types.NOT_PARSABLE.value
       self.val: str = ""
       self.name: str = ""
       self.var_type: str = ""
```
vvaltchev, line 41: **BAD_STYLE:** `var_type` should NOT be a string. Unless we're talking about very rare cases where there can be an unlimited amount of types coming from the outside world, types should be represented with Enums.

scripts/cmake-configurator/parser/cmake_row.py
```diff
@@ -32,33 +35,33 @@ def __init__(self, raw_row: str, row_number: int):
       """
       row  = raw_row.strip()
       matches: Optional[Match[str]]  = re.match(self.regex_expression, row)
-      self.type: str =  self.NOT_PARSABLE
+      self.type: str =  types.NOT_PARSABLE.value
       self.val: str = ""
```
vvaltchev, line 39: **ISSUE:** why `val` is a string? If can contain a `bool` or anything else and it's not serialized yet. Therefore, this is one of the few cases where you really shouldn't specify a type for the linter, unless there's something like `dynamic`, I don't remember. Also, initialize it to `None`, not to an empty string. Strings should be used only where it's strictly necessary. This is kind of the opposite of what happens in shell scripting (e.g. Bash) where everything is a string :-)

vvaltchev (changes requested):
Good work, moving in the direction of more type safety and fewer strings. 
Just, there are still additional things to do in this direction.

Also, can you rename (in this change or in a next one) the cmake-configurator directory to just `configurator`, please? The current name stops the auto-completion when trying to run the `cmake_run` script, which I run very often.

scripts/cmake-configurator/parser/cmake_row.py
```diff
@@ -1,5 +1,13 @@
 import re
 from typing import Match, Optional
+from enum import Enum
+
+class types(Enum):
```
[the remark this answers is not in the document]
michelececcacci: Changed to `row_types`

michelececcacci:
> Good work, moving in the direction of more type safety and fewer strings. Just, there are still additional things to do in this direction.
> 
> Also, can you rename (in this change or in a next one) the cmake-configurator directory to just `configurator`, please? The current name stops the auto-completion when trying to run the `cmake_run` script, which I run very often.

Changed to `configurator`

scripts/cmake-configurator/parser/cmake_row.py
michelececcacci: Done, now code uses enums properly

scripts/cmake-configurator/parser/cmake_row.py
michelececcacci: Wouldn't initialzing it to `None` cause more edge case handling?

scripts/cmake-configurator/parser/cmake_row.py
michelececcacci: Should we use strings here? it makes more sense for the final serialization

vvaltchev: Maybe I'm confused. What self.val is supposed to contain? I saw it might contain the actual bool value in case the type is BOOL, right? In case the type is something else, like integer, shouldn't it contain an integer instead of a string representing it? Of course if the type is STRING, then it will contain a string value. 

The idea of serialization is to convert the in-memory representation object to a flat-string. So, from self.val, we "compute" seralized_val, right? Actually my suggested code is wrong as well, because it takes self.val as it is in all the other cases and that doesn't look right.

scripts/cmake-configurator/parser/cmake_row.py
vvaltchev: It depends on what self.val is supposed to mean. If it's a in-memory representation of something we parsed (so, deserialized), it cannot be a string. It has to be a generic dynamic object. See the comments about the previous issue.

michelececcacci: Sorry, i got confused. I was thinking about line 38. You're totally right.

scripts/cmake-configurator/parser/cmake_row.py
michelececcacci: Yeah, that's right. Right now the only `var_type` that actually needs to change its `var_value` is `BOOL`, so i think your code would actually work, but it might need to be extended in the future. The only thing that needs to be changed from your code is  
```python
      if self.var_type == var_type.BOOL:
````

File: scripts/configurator/parser/cmake_row.py [hunk 7]
```diff
@@ -9,20 +23,22 @@ class cmake_row:
...
+   NORMALIZED_BOOLS = {
+      "ON" : True,
+      "OFF": False,
+      "1" : True,
+      "0" : False,
+      "TRUE": True,
+      "FALSE": False,
+   }
+
+   STR_TO_ENUM  = {
```
vvaltchev, line 37: Here this shouldn't be STR_TO_ENUM. This should be something like CMakeTypeToConfiguratorTypeMap.

scripts/configurator/parser/cmake_row.py
```diff
@@ -32,33 +48,33 @@ def __init__(self, raw_row: str, row_number: int):
       """
       row  = raw_row.strip()
       matches: Optional[Match[str]]  = re.match(self.regex_expression, row)
-      self.type: str =  self.NOT_PARSABLE
+      self.type = row_type.NOT_PARSABLE
       self.val: str = ""
       self.name: str = ""
       self.var_type: str = ""
```
vvaltchev, line 54: Why this is still a string?

scripts/configurator/parser/cmake_row.py
```diff
@@ -32,33 +48,33 @@ def __init__(self, raw_row: str, row_number: int):
...
       elif matches: # this guarantees that the regex has exactly 3 matches
          groups = matches.groups()
          self.name = groups[0]
-         self.type = self.VARIABLE
+         self.type = row_type.VARIABLE
          self.val = groups[2]
-         self.var_type = groups[1]
-         if self.var_type == "BOOL":
-            self.val = self.BOOL_MAPPINGS[self.val]
+         self.var_type = self.STR_TO_ENUM[groups[1]]
```
vvaltchev, line 75: And here you're treating it as an object. Did you run mypy? This shouldn't pass.

scripts/configurator/parser/cmake_row.py
```diff
@@ -1,5 +1,19 @@
...
+
+class row_type(Enum):
+   EMPTY = auto()
+   SLASH_COMMENT = auto()
+   HASHTAG_COMMENT = auto()
+   VARIABLE = auto()
+   NOT_PARSABLE = auto()
+
+class var_type(Enum):
+   BOOL = "BOOL"
```
vvaltchev, line 13: I really HATE this. This is Python, there must be a better way. But, before getting there: the first point here is asking ourselves which is fixed as does not depend on us? I guess the CMake string types. Therefore, we have to keep them, but can change the rest (left side).

Now, why are we declaring the same types here? We're supposed to support a super set of those types and there won't be a bi-directional 1:1 mapping. Each of our types, will map to exactly one of those CMake types, but the opposite isn't true! Therefore, let's define a list of our types, using a different case, names etc. and avoid confusion with CMake types. From this code, it should be clear that those are two separate sets of types which have a UNIDIRECTIONAL mapping 1:1. In the opposite direction, the CMake type STRING will have make many our types mapped to, like INT, BASE_2_INT, etc.

vvaltchev (changes requested):

scripts/configurator/parser/cmake_row.py
```diff
@@ -75,7 +91,13 @@ def serialize(self) -> str:
       returns serialized version of variable. Expects to be called only on
       variable types.
       """
-      return self.name + ":" + self.var_type + "=" + self.val
+      serialized_val = self.val
```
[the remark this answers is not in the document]
michelececcacci: That's right, i staged the commit but didn't push. sorry.

scripts/configurator/parser/cmake_row.py
michelececcacci: You're right. I will fix this

scripts/configurator/parser/cmake_row.py
michelececcacci: Changed.

scripts/configurator/parser/cmake_row.py
```diff
@@ -49,9 +49,9 @@ def __init__(self, raw_row: str, row_number: int):
       row  = raw_row.strip()
       matches: Optional[Match[str]]  = re.match(self.regex_expression, row)
       self.type = row_type.NOT_PARSABLE
-      self.val: str = ""
+      self.val: str | bool = ""
```
vvaltchev, line 52: I appreciate the effort here. That certainly is way better than before, but I still don't like the initialization to empty string. Empty scripts used as a marker for unset is a very bad **anti-pattern**: please-don't do that. Initialize to `None` and deal with that.

About the type, what I really wanted was `Any` because that's what `val` is (it could be anything) but.. if it's possible to specify a list of types instead (as you're doing), I guess that could work as well.

scripts/configurator/parser/cmake_row.py
```diff
@@ -91,10 +91,13 @@ def serialize(self) -> str:
       returns serialized version of variable. Expects to be called only on
       variable types.
       """
-      serialized_val = self.val
+      serialized_val: str
+      # only way to to shut mypy up
```
vvaltchev, line 95: yep, and mypy is right ;-)

vvaltchev (changes requested):
Could you please squash? I'm a bit confused at this point and wanted to see in the easiest way the final shape of the code, in particular things around var_type vs row_type etc.

scripts/configurator/parser/cmake_row.py
michelececcacci: > Yeah, sure, i am not done with modifications yet, but i'll squash

scripts/configurator/parser/cmake_row.py
michelececcacci: I could create a composite type, if you feel like that would reflect better the idea we are trying to convey

scripts/configurator/parser/cmake_row.py
michelececcacci: > Sure, i will squash, but this is not the final shape of the code yet. I will ask for review again when it's actually ready to be reviewed.

scripts/configurator/parser/cmake_parser.py
```diff
@@ -0,0 +1,37 @@
+from cmake_row import cmake_row, configurator_type, row_type
+from typing import List
+
+def read_file(filepath):
+   with open(filepath) as f:
+      return f.readlines()
+
+def lines_to_rows(file_lines: List[str]) -> List[cmake_row]:
+   """
+   turns the lines of a file (parsed previously) to a list of cmake_rows,
```
vvaltchev, line 10: MISLEADING COMMENT: the parsing is what happens where, previously we just split the file in lines

scripts/configurator/parser/cmake_parser.py
```diff
@@ -0,0 +1,37 @@
...
+   cmake_rows: List[cmake_row] =  []
+   comment = ""
+
+   for i, line in enumerate(file_lines):
+      row = cmake_row(line, i, comment)
+      if row.row_type == row_type.SLASH_COMMENT:
+         comment += row.val
+      elif row.row_type == row_type.VARIABLE and \
+            row.configurator_type != configurator_type.INTERNAL:
+         cmake_rows.append(row)
```
vvaltchev, line 22: 1. why are you appending to rows only the variables? I believe we talked about this: in this approach ALL the lines got converted to cmake_row instances.

2. you're appending too many comments to variables. This approach is not OK. Either you walk back the comments once if found a variable or you walk always forward but reset the comment when you hit an empty line. Btw, the second approach is less intuitive to both read & write.

scripts/configurator/parser/cmake_parser.py
```diff
@@ -0,0 +1,37 @@
...
+         comment = ""
+
+      return cmake_rows
+
+def rows_to_lines(file_lines: List[str],
+                  cmake_rows: List[cmake_row]) -> List[str]:
+   for row in cmake_rows:
+      if row.row_type == row_type.VARIABLE:
+         file_lines[row.row_number] = row.serialize()
```
vvaltchev, line 31: **NOPE!** Yesterday I wrote the pseudo code just to avoid this mistake once again:
I'm copy-pasting this for the last time. The with the rows + variables approach, the serialize should look like this:
```
for each row in rows:
    fd.write(row.serialize() + "\n")
```

We talked about both the approaches. If you want to keep the rows, you discard the file lines. You have **only** rows now! The the alternative is that you have only file lines and variables, but no rows. In that case you iterate over all the variables and replace the right lines with the newly serialized value.

It should be simple: 
de-serialize: input=lines, output=cmake_rows
serialize: input=cmake_rows, output=lines

The row.serialize() method should know what to do in case the row contains a variable and call self.var.serialize() or something like that.

vvaltchev (changes requested):
Please squash everything. Also, please change the commit message to something like "global refactoring".

scripts/configurator/parser/cmake_parser.py
michelececcacci: Fixed.

scripts/configurator/parser/cmake_row.py
```diff
@@ -0,0 +1,105 @@
...
+
+   def __init__(self, raw_row: str, row_number: int, comment: str = ""):
+      """
+      strips variable and assigns its correct type by matching the regex
+      and verifying if it's a comment.
+      """
+      row  = raw_row.strip()
+      matches: Optional[Match[str]]  = re.match(self.regex_expression, row)
+      self.row_type = row_type.NOT_PARSABLE
+      self.val: str | bool = ""
```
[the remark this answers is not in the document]
vvaltchev: Also, if you re-add the cmake_variable class, this should become something like:
`self.val : str | bool | cmake_variable = None`

scripts/configurator/parser/cmake_row.py
```diff
@@ -0,0 +1,105 @@
...
+         return
+
+      elif matches: # this guarantees that the regex has exactly 3 matches
+         groups = matches.groups()
+         self.name = groups[0]
+         self.row_type = row_type.VARIABLE
+         self.val = groups[2]
+         self._cmake_type_str = groups[1]
+         self.configurator_type = self.string_to_conf_type[self._cmake_type_str]
+         if self.configurator_type == configurator_type.BOOL:
```
vvaltchev, line 77: That's the place where you should create a cmake_variable can give it the string-representation in order to the variable class to de-serialize it, instead of doing it everything in the `cmake_row` class.

scripts/configurator/parser/cmake_row.py
```diff
@@ -0,0 +1,105 @@
...
+         else:
+            serialized_val = str(self.val)
+         return self.name + ":" + self._cmake_type_str + "="  + serialized_val
+      elif self.row_type == row_type.POUND_COMMENT:
+         return "#" + self.val
+      elif self.row_type == row_type.SLASH_COMMENT:
+         return "//" + self.val
+      return "" # in case of empty string
+
+   def normalize_bool(self, val: str) -> bool:
```
vvaltchev, line 104: This as the NORMALIZED_BOOLS global can be in the cmake_variable file.

vvaltchev (changes requested):
Thanks for the update! It's much better than before. 

Still, there are things requiring fixing. For example the cmake_variable would be very useful here. We agreed to remove the file_lines, not to remove the variable as well. It's an useful abstraction, see all the places (in the comments) where it would be nicer to have it.

scripts/configurator/parser/test_cmakeparser.py
```diff
@@ -1,5 +1,6 @@
-from cmake_parser import cmake_parser
+from re import L
 from unittest import TestCase
+from cmake_parser import lines_to_rows, configurator_type, write_file, row_type
```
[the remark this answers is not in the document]
michelececcacci: PEP 8 Heavily discourages that, because it's not esplicit and it can lead to circular imports

scripts/configurator/parser/cmake_row.py
```diff
@@ -0,0 +1,105 @@
...
+   NORMALIZED_BOOLS = {
+      "ON" : True,
+      "OFF": False,
+      "1" : True,
+      "0" : False,
+      "TRUE": True,
+      "FALSE": False,
+   }
+
+   string_to_conf_type  = {
```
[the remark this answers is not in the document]
michelececcacci: Isn't `string` more explicit? i feel like this implies that we use another enum called `cmake_type`

scripts/configurator/parser/test_cmakeparser.py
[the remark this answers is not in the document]
vvaltchev: Well, I believe that applying blindly even very good coding standards is not the best. `from X import *` can cause problems indeed, but makes sense when you're literally importing _every_ single function and class from a module, especially for simple and small modules like that one. It's not like doing `using namespace std;` in C++. It's like using your local mini-namespace defined in a another file in the same directory. Indeed, for small C++ projects we use the same namespace for all the files. Just, in this case Python automatically creates a new namespace for each file. And... about the cycling dependencies: I actually appreciate the fact that `import *` will fail in this case, because it will force me to avoid the cyclic dependency. When possible, it's _much better_ to avoid having 2 files each one of them depending on the other. When that happens, it is a _strong sign_ that maybe those file should be just merged into one, as the coupling is too tight for them to be separate.

Also, the scale matters: what is bad practice for the medium-large scale, can be perfectly acceptable or even _good_ for the small-scale projects. The [FizzBuzzEnterpriseEdition](https://github.com/EnterpriseQualityCoding/FizzBuzzEnterpriseEdition) project is a good example of what happens when some mostly good coding practices used in enterprise large-scale are forced in small-scale project. 

Still, I won't insist further on this point, as there is no damage in the extra verbosity. Feel free to keep the explicit list of imports. I just hope that I made you at least think more about the problem and all the implications. Coding standard rules are generally not written in stone, but are _guidelines_ generally designed for medium-large projects.

scripts/configurator/parser/cmake_parser.py
```diff
@@ -5,7 +5,7 @@ def read_file(filepath):
    with open(filepath, 'r') as f:
       return f.readlines()
 
-def lines_to_rows(file_lines: List[str]) -> List[cmake_row]:
+def parse_variables(file_lines: List[str]) -> List[cmake_row]:
```
vvaltchev, line 8: `parse_lines` would be better. We're creating row objects as a results, not just variables.

scripts/configurator/parser/cmake_row.py
```diff
@@ -45,8 +44,8 @@ def __init__(self, raw_row: str, row_number: int, comment: str = ""):
       """
       row  = raw_row.strip()
       matches: Optional[Match[str]]  = re.match(self.regex_expression, row)
-      self.row_type = row_type.NOT_PARSABLE
-      self.val: str | bool = ""
+      self.row_type: row_type | None = None
```
vvaltchev, line 47: Good but where do we fail? At the end of the `if/elif` sequence there must be an `else: raise RuntimeError("...")`. We cannot just silently leave the `row_type` as None.

vvaltchev (changes requested):
Minor comments. Keep in mind that the re-addition of cmake_var & related changes can be done in another pull request. So, if you fix the minor details we could merge this as well.

scripts/configurator/parser/cmake_parser.py
vvaltchev: Or `parse_rows` seems good as well. Try to thing very carefully what's the input and what's the output. The name should be as descriptive as possible.

scripts/configurator/parser/cmake_row.py
```diff
@@ -0,0 +1,104 @@
...
+      self.name: str = ""
+      self._cmake_type_str: str # used to serialize back the value
+      self.configurator_type: configurator_type
+      self.row_number: int = row_number
+      self.comment: str = comment
+
+      if not len(row):
+         self.row_type = row_type.EMPTY
+
+      if self.is_slash_comment(row):
```
vvaltchev, line 58: **FIX**: this should be an `elif`

vvaltchev (changes requested):
One `elif` is missing. Otherwise is fine.

scripts/configurator/parser/cmake_row.py
michelececcacci: Changed

PR #123, opened 2022-08-09
[configurator] extracted rendundant functionalities into cmake_var

michelececcacci:
Referencing #114

### DIFF
--- a/scripts/configurator/parser/cmake_row.py
+++ b/scripts/configurator/parser/cmake_row.py
@@ -1,18 +1,13 @@
 import re
 from typing import Match, Optional
 from enum import Enum, auto
+from cmake_var import cmake_var, build_cmake_var
 
 class row_type(Enum):
    EMPTY = auto()
    SLASH_COMMENT = auto()
    POUND_COMMENT = auto()
    VARIABLE = auto()
-
-class configurator_type(Enum):
-   BOOL = auto()
-   FILEPATH = auto()
-   STRING = auto()
-   INTERNAL = auto()
 
 class cmake_row:
    """
@@ -21,39 +16,22 @@ class cmake_row:
    """
    regex_expression = re.compile(r"^([A-Za-z_0-9]+?):([A-Za-z]+?)=(.*)$")
 
-   NORMALIZED_BOOLS = {
-      "ON" : True,
-      "OFF": False,
-      "1" : True,
-      "0" : False,
-      "TRUE": True,
-      "FALSE": False,
-   }
-
-   cmake_type_to_conf_type  = {
-      "BOOL": configurator_type.BOOL,
-      "FILEPATH": configurator_type.FILEPATH,
-      "STRING": configurator_type.STRING,
-      "INTERNAL": configurator_type.INTERNAL,
-   }
-
    def __init__(self, raw_row: str, row_number: int, comment: str = ""):
       """
       strips variable and assigns its correct type by matching the regex
       and verifying if it's a comment.
       """
       row  = raw_row.strip()
-      variable_match: Optional[Match[str]]  = re.match(self.regex_expression, row)
-      self.row_type: row_type | None = None
-      self.val: str | bool | None = None
+      variable_match: Optional[Match[str]] = re.match(self.regex_expression, row)
+      self.val: str | cmake_var | None = None
       self.name: str = ""
-      self._cmake_type_str: str # used to serialize back the value
-      self.configurator_type: configurator_type
+      self._cmake_type: str # used to serialize back the value
       self.row_number: int = row_number
       self.comment: str = comment
 
       if not len(row):
          self.row_type = row_type.EMPTY
+         self.val = ""
 
       elif self.is_slash_comment(row):
          self.row_type = row_type.SLASH_COMMENT
@@ -67,12 +45,10 @@ class cmake_row:
          groups = variable_match.groups()
          self.name = groups[0]
          self.row_type = row_type.VARIABLE
-         self.val = groups[2]
-         self._cmake_type_str = groups[1]
-         self.configurator_type = \
-            self.cmake_type_to_conf_type[self._cmake_type_str]
-         if self.configurator_type == configurator_type.BOOL:
-            self.val = self.normalize_bool(self.val)
+         val = groups[2]
+         self._cmake_type = groups[1]
+         self.val = build_cmake_var(self._cmake_type, val)
+
       else:
          raise ValueError("Could not parse variable")
 
@@ -87,18 +63,19 @@ class cmake_row:
       return row[0] == "#"
 
    def serialize(self) -> str:
-      serialized_val: str
       if self.row_type == row_type.VARIABLE:
-         if self.configurator_type == configurator_type.BOOL:
-            serialized_val = "ON" if self.val else "OFF"
-         else:
-            serialized_val = str(self.val)
-         return self.name + ":" + self._cmake_type_str + "="  + serialized_val
+         return self.name + ":" + self._cmake_type + "=" + self.val.serialize()
       elif self.row_type == row_type.POUND_COMMENT:
-         return "#" + self.val
+         return "#" + str(self.val)
       elif self.row_type == row_type.SLASH_COMMENT:
-         return "//" + self.val
-      return "" # in case of empty string
+         return "//" + str(self.val)
+      elif self.row_type == row_type.EMPTY:
+         return ""
+      else:
+         raise ValueError("Row does not have a defined type")
 
-   def normalize_bool(self, val: str) -> bool:
-      return self.NORMALIZED_BOOLS[val]
+   def get_val(self):
+      if self.row_type == row_type.VARIABLE:
+         return  self.val.serialize()
+      else:
+         return self.val
--- a/scripts/configurator/parser/cmake_var.py
+++ b/scripts/configurator/parser/cmake_var.py
@@ -1,0 +1,38 @@
+from typing import Dict, Any
+
+def build_cmake_var(type :str, cmake_var_str :str):
+   "factory function that associates the correct type of variable"
+   cmake_type_str_to_class = {
+      "BOOL" : cmake_var_bool,
+      "INTERNAL" : cmake_var,
+      "FILEPATH": cmake_var,
+      "STRING" : cmake_var,
+   }
+   class_obj = cmake_type_str_to_class[type]
+   return class_obj(cmake_var_str)
+
+
+class cmake_var:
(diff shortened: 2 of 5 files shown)


scripts/configurator/parser/cmake_var.py
```diff
@@ -0,0 +1,39 @@
...
+      "BOOL" : cmake_var_bool,
+      "INTERNAL" : cmake_var,
+      "FILEPATH": cmake_var,
+      "STRING" : cmake_var,
+   }
+   variable_instance = string_to_variable_class[type]
+   return variable_instance(var)
+
+
+class cmake_var:
```
vvaltchev, line 13: This should be something like `cmake_var_str`.

scripts/configurator/parser/cmake_var.py
```diff
@@ -0,0 +1,39 @@
...
+   "generic type of cmake_var, used when we don't need particular conversions"
+   def __init__(self, var: str) -> None:
```
vvaltchev, line 15: `var` should be `val`. The variable has a _value_ so we're now dealing with it.

scripts/configurator/parser/cmake_var.py
```diff
@@ -0,0 +1,39 @@
+def generate_cmake_var(type :str, var):
+   "factory function that associates the correct type of variable"
+   string_to_variable_class = {
```
vvaltchev, line 3: Rename to `cmake_type_str_to_class`

scripts/configurator/parser/cmake_var.py
```diff
@@ -0,0 +1,39 @@
...
+      "BOOL" : cmake_var_bool,
+      "INTERNAL" : cmake_var,
+      "FILEPATH": cmake_var,
+      "STRING" : cmake_var,
+   }
+   variable_instance = string_to_variable_class[type]
```
vvaltchev, line 9: NIT: No need for a variable (too long), but also, `variable_instance` is wrong: you're getting a `class` object, not an instance. So, at most it should be something like `class_obj`. Later, you're creating an object instance from the class.

vvaltchev (changes requested):
Good change! There are a few mostly cosmetic changes to make.
Btw, you didn't use inheritance and classic polymorphism. Now the cmake_var_* classes are completely unrelated. That's totally fine in a dynamic language like Python, but it won't work in other languages like Java. Still, again, no need to forcibly use inheritance here, I'm happy with the current code. I just wanted to point out this aspect.

scripts/configurator/parser/cmake_var.py
michelececcacci: in place of `var`, right?

scripts/configurator/parser/cmake_var.py
michelececcacci: Ok, but since when we get the  val  from `cmake_row` it can effectively be either the value (in case of comments) or the `cmake_var`  holding it, i think it's just better to insert a setter to get the value no matter what.

scripts/configurator/parser/cmake_var.py
vvaltchev: `cmake_var` should be renamed to `cmake_var_str` or, if you want a generic no-op type, you could name it `cmake_var_generic`.

scripts/configurator/parser/cmake_var.py
vvaltchev: No, wait. I've been talking about the `var` parameter and `var` member in `cmake_var`. Not the "_var" suffix in the same. The variable (aka cmake_var) has a value. You can call it also `value` in order everything to be even clearer.

scripts/configurator/parser/cmake_var.py
michelececcacci: Ok. I think switching names between union type and class name is more appropriate

scripts/configurator/parser/cmake_var.py
michelececcacci: Ok, makes sense. What about the getter idea? do you like it?  I think makes sense

scripts/configurator/parser/cmake_var.py
```diff
@@ -0,0 +1,40 @@
...
+      "BOOL" : cmake_var_bool,
+      "INTERNAL" : cmake_var_generic,
+      "FILEPATH": cmake_var_generic,
+      "STRING" : cmake_var_generic,
+   }
+   class_obj = cmake_type_str_to_class[type]
+   return class_obj(cmake_var_str)
+
+
+class cmake_var_generic:
```
vvaltchev, line 15: Sorry for the confusion, in case I haven't been clear enough. I meant to use `cmake_var_generic` if you didn't want to have a base class + inheritance, because that's why I saw in the code and I told you it was fine, even if it wouldn't work on other languages.

Now that you made `cmake_var_bool` to inherit from `cmake_var_generic`, the name `cmake_var_generic` is not good enough. It has to be `cmake_var` like you previously named it or `cmake_var_base` at most. 

In conclusion, either you call it `cmake_var_generic` but don't use inheritance, or you call it `cmake_var` and you do use inheritance. The previous change was a mix (`cmake_var` but without inheritance) and was semantically weird.

scripts/configurator/parser/cmake_var.py
```diff
@@ -0,0 +1,40 @@
...
+      "FALSE": False,
+   }
+
+   def __init__(self, value: str) -> None:
+      self.value = self.NORMALIZED_BOOLS[value]
+
+   def serialize(self) -> str:
+      return "ON" if self.value else "OFF"
+
+cmake_var = cmake_var_generic | cmake_var_bool
```
vvaltchev, line 40: Can you explain to be how this syntax works? It doesn't look like a linter thing. It looks like a type alternative that I saw only in pure functional languages like Haskell. I didn't know you could do things like that in Python.

Btw, if you use inheritance, you don't use type alternatives like that. You just use the base class, which should be simply called `cmake_var`. That's the whole point of polymorphism. Type alternatives are good when the types are NOT related. It's either one or the other. It doesn't make sense like that.

scripts/configurator/parser/cmake_var.py
```diff
@@ -0,0 +1,40 @@
+from typing import Dict, Any
+
+def generate_cmake_var(type :str, cmake_var_str):
```
vvaltchev, line 3: NAMING: This is a factory function, not a generator of any kind. Please don't use the `generate_` prefix. Typically `make` or `build` are used in the industry.

vvaltchev (changes requested):
Very good! There are just a few additional renamings to do related with the previous discussions and one new.

scripts/configurator/parser/cmake_var.py
michelececcacci: Ok, that makes sense. `X | Y` is just syntactic sugar for `Union[X, Y]`. I think it's also used in typescript.

scripts/configurator/parser/cmake_row.py
```diff
@@ -87,18 +62,16 @@ def is_pound_comment(self, row: str) -> bool:
...
-      serialized_val: str
       if self.row_type == row_type.VARIABLE:
-         if self.configurator_type == configurator_type.BOOL:
-            serialized_val = "ON" if self.val else "OFF"
-         else:
-            serialized_val = str(self.val)
-         return self.name + ":" + self._cmake_type_str + "="  + serialized_val
-      elif self.row_type == row_type.POUND_COMMENT:
+         return self.name + ":" + self._cmake_type + "=" + self.val.serialize()
+      elif type(self.val) == str and self.row_type == row_type.POUND_COMMENT:
```
vvaltchev, line 67: No need to check for `type(self.val) == str` here. This is an invariant of your code and always must hold when the row_type is a comment. Therefore, just check for `self.row_type == row_type.POUND_COMMENT` and, if you want, just `assert(type(self.val) == str)` inside the `if`. With `if` statements we should check for things that are expected to be true or false. Invariants are checked with asserts, instead.

Also, the whole serialize() should be conceptually implemented with a big switch statement. Because python doesn't have that, a sequence of if/elif will do the job.

scripts/configurator/parser/cmake_row.py
```diff
@@ -87,18 +62,16 @@ def is_pound_comment(self, row: str) -> bool:
...
          return "#" + self.val
-      elif self.row_type == row_type.SLASH_COMMENT:
+      elif type(self.val) == str and self.row_type == row_type.SLASH_COMMENT:
```
vvaltchev, line 69: Same as before.

NIT: It would be more polished to have an `elif` case for the empty string and a final `else` statement where you throw an exception. In other words, the `if/elif` list here must be _exhaustive_ and cover all the types that `row` could possibly have. In the current implementation, you simply assume that there are no more types to cover and default to empty string case: that is correct (obviously) in this case, but "doesn't scale" in large projects with many types to cover. The cool thing about C and C++ switch statements is that, if you do it on an enum type, the compiler will show you a warning in case you missed a case. Anyway, this is an optional NIT as we don't have that scale here and `row` won't have many more types.

scripts/configurator/parser/cmake_row.py
```diff
@@ -87,18 +62,16 @@ def is_pound_comment(self, row: str) -> bool:
...
          return "//" + self.val
       return "" # in case of empty string
 
-   def normalize_bool(self, val: str) -> bool:
-      return self.NORMALIZED_BOOLS[val]
+   def get_val(self):
```
vvaltchev, line 73: I get what you'd like to have with this `get_val()`. I have a bit of mixed feelings, but I tend for keeping it at the moment, as it does a great job for the tests.

vvaltchev (changes requested):
Almost perfect, very few cosmetic comments.

scripts/configurator/parser/cmake_row.py
michelececcacci: This is just needed for mypy to stop complaining. The invariant holds, but unfortunately mypy doesn't like it.

scripts/configurator/parser/cmake_row.py
michelececcacci: Ok, this makes sense

scripts/configurator/parser/cmake_row.py
michelececcacci: Ok, perfect

scripts/configurator/parser/cmake_row.py
vvaltchev: OK, then use `str(self.val)` instead of this ugly checking.

PR #124, opened 2022-08-10
[configurator] using ABC for cmake_var

michelececcacci:
Using an ABC instead of simply derivating. I feel like it reflects better what we are trying to do

### DIFF
--- a/scripts/configurator/parser/cmake_row.py
+++ b/scripts/configurator/parser/cmake_row.py
@@ -1,5 +1,5 @@
 import re
-from typing import Match, Optional
+from typing import Match, Optional, Type
 from enum import Enum, auto
 from cmake_var import cmake_var, build_cmake_var
 
@@ -28,6 +28,7 @@ class cmake_row:
       self._cmake_type: str # used to serialize back the value
       self.row_number: int = row_number
       self.comment: str = comment
+      self.row_type: row_type
 
       if not len(row):
          self.row_type = row_type.EMPTY
--- a/scripts/configurator/parser/cmake_var.py
+++ b/scripts/configurator/parser/cmake_var.py
@@ -1,18 +1,16 @@
-from typing import Dict, Any
+from typing import Dict, Any, Mapping, Type
+from abc import ABCMeta, abstractmethod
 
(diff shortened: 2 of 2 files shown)


vvaltchev (changes requested):
I'm sorry if that might disappoint you, but this is textbook overengineering. The fact that ABCMeta exists, doesn't mean we have to use it for first trivial inheritance we introduced. Such libraries have been introduced to handle way more complicated cases in big projects. We have just one base class and two derived classes and you're already trying to most powerful tool available. It feels like when I used C++ metaprogramming techniques to solve minor problems just because it was cool :-)  

So, please: _do not use a cannon to kill a mosquito_ (Confucius)

michelececcacci:
> I'm sorry if that might disappoint you, but this is textbook overengineering. The fact that ABCMeta exists, doesn't mean we have to use it for first trivial inheritance we introduced. Such libraries have been introduced to handle way more complicated cases in big projects. We have just one base class and two derived classes and you're already trying to most powerful tool available. It feels like when I used C++ metaprogramming techniques to solve minor problems just because it was cool :-)
> 
> So, please: _do not use a cannon to kill a mosquito_ (Confucius)

Should i use a normal ABC instead?

vvaltchev:
@michelececcacci Sorry, but I genuinely believe the whole change should be dropped, in the name of the KISS principle. There is no real need to use the `abc` module at all, at the moment. I know it's cool and powerful, but it's also a huge overkill in this trivial case. Let's use it in case the complexity of the configurator tool raises significantly.

Again, don't get me wrong: I'm sure that with `abc` we could do powerful OOP, but do we really need that now? What we're gaining from it, in the current context? We're adding an unnecessary dependency to a fancy module (standard module, I know, but still) and we're just adding more syntax sugar. Using simple inheritance was reasonably the most "powerful solution" I was willing to accept and it already was a bit over-engineering. Declaring a whole abstract class I way too much.

A more pythonic solution would be do use a dict with `serialize()` functions, no classes, no objects :-) Doing every possible thing with OOP is overrated today.

michelececcacci:
Ok, i think you are right.

PR #128, opened 2022-08-19
[configurator] Added more type checking

michelececcacci:
Behaviour is the same as before, but code is more explicit


vvaltchev (approved):
The change looks good to me :-)

PR #130, opened 2022-08-24
Configurator: add a unit test for the write_file() func + fixes in the CI

vvaltchev:
@michelececcacci


michelececcacci:
I really like this commit. Just a minor thing. Why are we making configurator be triggered by just paths?

michelececcacci:
LGTM

PR #131, opened 2022-08-26
[configurator] adding metadata support

michelececcacci:


scripts/configurator/parser/cmake_var.py
```diff
@@ -1,7 +1,104 @@
...
+      self.kw_to_parse: Mapping[str, function] = {
+         "REALTYPE" : self.parse_realtype,
+         "DEPENDS_ON" : self.parse_depends_on,
+         "GROUP" : self.parse_group
+      }
+
+      self.kw_to_setter = {
+         "REALTYPE": self.get_setter_realtype,
+         "DEPENDS_ON": self.get_setter_depends_on,
+         "GROUP": self.get_setter_group,
```
vvaltchev, line 23: This is super ugly. Do:

```
      self.kw_to_setter = {
         "REALTYPE": self.get_setter_realtype(),
         "DEPENDS_ON": self.get_setter_depends_on(),
         "GROUP": self.get_setter_group(),
    }
```

you really need to put the **setter closure** in that dict, **not** the getter function for the closure.

scripts/configurator/parser/cmake_var.py
```diff
@@ -1,7 +1,104 @@
...
+               continue
+            elif key not in self.kw_to_parse:
+               msg = "{} is not a valid metadata keyword."
+               raise ValueError(msg.format(key))
+            elif self.kw_to_var[key] is not None:
+               msg = "Key {} has already a value associated to itself."
+               raise ValueError(msg.format(key))
+            parse_function = self.kw_to_parse[key]
+            not_matched = line[len(key):].lstrip()
+            get_setter= self.kw_to_setter[key]
```
vvaltchev, line 48: If you put the closures in the dict, as I suggested above, here you'll just do:

```
setter = self.kw_to_setter[key]
```

vvaltchev (changes requested):
This is not a full review, just the first things I noticed.

scripts/configurator/parser/cmake_parser.py
```diff
@@ -14,9 +14,15 @@ def parse_rows(file_lines: List[str]) -> List[cmake_row]:
       if current_row.row_type == row_type.VARIABLE:
          j = i-1
          while j >= 0 and cmake_rows[j].row_type == row_type.SLASH_COMMENT:
-            comment = cmake_rows[j].get_val() + comment
+            comment = cmake_rows[j].get_val() + "\n" + comment
```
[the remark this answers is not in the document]
michelececcacci: I thought about it, and i think it's the best thing to do, but at the same time i feel like a string conveys the data type better. i can definitely do it, though.

scripts/configurator/parser/cmake_var.py
michelececcacci: oh, ok. That actually makes sense

scripts/configurator/parser/cmake_var.py
michelececcacci: perfect. yes, that's defintely better

scripts/configurator/parser/cmake_parser.py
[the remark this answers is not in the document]
michelececcacci: Since we are appending at the beginning, should i use a deque to append in O(1)?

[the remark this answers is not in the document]
vvaltchev: Even if at this scale doesn't matter, it's a good point. Btw, the prepend operation for the string was O(n) when n = total num of characters, while for the python list will be O(n) where n = number of lines, not chars. Therefore, by just converting `comment` to `comment_lines` we would significantly improve this code.

Still, using deque would be fine. But I guess that keeping `comment_lines` as a list is more convenient in the long run as we won't need to manipulate it like a deque and primitive types should be generally preferred to library ones. So, you could create a temp. deque and convert it to a list at the end.

scripts/configurator/parser/cmake_parser.py
```diff
@@ -1,23 +1,30 @@
...
-   comment = ""
+   comment_list: Deque[str] = deque()
    for i, line in enumerate(file_lines):
       current_row = cmake_row(line, i)
       cmake_rows.append(current_row)
       if current_row.row_type == row_type.VARIABLE:
          j = i-1
          while j >= 0 and cmake_rows[j].row_type == row_type.SLASH_COMMENT:
-            comment = cmake_rows[j].get_val() + comment
+            comment_list.appendleft(cmake_rows[j].get_val() + "\n")
```
vvaltchev, line 18: **BUG:** now that we have a list of comment rows, it doesn't make any sense to keep "\n".

vvaltchev (changes requested):

scripts/configurator/parser/cmake_parser.py
michelececcacci: That's actually true, you're totally right

scripts/configurator/parser/cmake_parser.py
```diff
@@ -1,23 +1,30 @@
...
             j -= 1
-         current_row.comment = comment
-         comment = ""
+         """
+         In this case we need a new cmake_row to handle all the new
+         information provided by the metadata, so every cmake_row will be
+         constructed twice. The process could also be extracted to other
+         functions, but this respects the seperation of concerns.
+         """
+         cmake_rows[i] = cmake_row(line, j, comment_list)
```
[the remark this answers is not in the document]
michelececcacci: Ok, that makes sense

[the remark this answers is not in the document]
michelececcacci: I'm not sure if i understand it correctly. My approach would be of just using a deque converting to list when needed (with `list(comment_list)`). is this correct?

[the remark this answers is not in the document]
vvaltchev: No, not when needed. Convert to list just once, before passing it to `cmake_row` and keep it as list.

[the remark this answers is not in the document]
michelececcacci: Wait, why do we care about keeping the list? it's the only case in which it will be used, in the other case we just use the default parameter, which is set to empty.

scripts/configurator/parser/cmake_var.py
```diff
@@ -1,9 +1,108 @@
...
+            elif self.kw_to_var[key] is not None:
+               msg = "Key {} has already a value associated to itself."
+               raise ValueError(msg.format(key))
+            parse_function = self.kw_to_parse[key]
+            not_matched = line[len(key):].lstrip()
+            setter = self.kw_to_setter[key]
+            s = parse_function(not_matched)
+            setter(s)
+
+   def get_setter_depends_on(self):
```
vvaltchev, line 54: After some thinking, I believe the amount of boilerplate code introduce by this setters is way too high, compared to a trivial if/elseif list. Therefore, we should really use just one generic setter like this:

```python

def get_setter(self, var_name):
   def setter(val):
      nonlocal self
      self.__dict__[var_name] = val
    return setter
```

scripts/configurator/parser/cmake_var.py
```diff
@@ -1,9 +1,108 @@
...
+
+class metadata:
+   keyword_expr = re.compile(r"([A-Z_]+)")
+
+   def __init__(self, comment: List[str]) -> None:
+      self.depends_on: Optional[List[str]] = None
+      self.group: Optional[str] = None
+      self.realtype: Optional[str] = None
+
+      self.kw_to_parse: Mapping[str, function] = {
```
vvaltchev, line 14: 3 dictionaries with the same key type are too many and that's a sign that something is wrong. In this case we should have just _one_ dictionary KEYWORD -> value with value being a tuple (e.g. namedTuple) containing: getter, setter, parser. This will make this whole code way shorter.

scripts/configurator/parser/cmake_var.py
```diff
@@ -1,9 +1,108 @@
...
+         "REALTYPE" : self.parse_realtype,
+         "DEPENDS_ON" : self.parse_depends_on,
+         "GROUP" : self.parse_group
+      }
+
+      self.kw_to_setter = {
+         "REALTYPE": self.get_setter_realtype(),
```
vvaltchev, line 21: with the generic setter, we'll have to pass here the name of the variable, but that's a good enough trade-off to avoid the boilerplate setter for each metadata field.

scripts/configurator/parser/cmake_var.py
```diff
@@ -1,9 +1,108 @@
...
+
+      self.kw_to_var = {
+         "REALTYPE": self.realtype,
+         "DEPENDS_ON": self.depends_on,
+         "GROUP": self.group,
+      }
+
+      for i, line in enumerate(comment):
+         match = self.keyword_expr.match(line)
+         if match:
```
vvaltchev, line 34: This `if` contains too much stuff and it's hard to read. Please use:
```
if not match:
   continue
```

scripts/configurator/parser/cmake_var.py
```diff
@@ -1,9 +1,108 @@
-from typing import Dict, Any, Type, Mapping
+from typing import Dict, Any, Optional, Type, Mapping, List
+import re
+from sys import stderr
+
+
+class metadata:
+   keyword_expr = re.compile(r"([A-Z_]+)")
```
vvaltchev, line 7: I didn't think we need to use a regex for the keywords. In the simple if/elif impl. I would have used `if line.startswith("GROUP ")` etc. But, with this implementation this is understandable. Just make it be: `^([A-Z_]+)` so that's more explicit we're requiring that at the beginning of the line.

scripts/configurator/parser/cmake_var.py
```diff
@@ -1,9 +1,108 @@
...
+
+   def __init__(self, comment: List[str]) -> None:
+      self.depends_on: Optional[List[str]] = None
```
vvaltchev, line 10: In a follow-up change, we'll have to check somewhere that the list in `depends_on` consists in actual bool variables. If the variables do not exist or they are not booleans, we have to show an error and fail.

scripts/configurator/parser/cmake_var.py
```diff
@@ -1,9 +1,108 @@
...
+                  stderr.write(msg.format(key))
+               continue
+            elif key not in self.kw_to_parse:
+               msg = "{} is not a valid metadata keyword."
+               raise ValueError(msg.format(key))
+            elif self.kw_to_var[key] is not None:
+               msg = "Key {} has already a value associated to itself."
+               raise ValueError(msg.format(key))
+            parse_function = self.kw_to_parse[key]
+            not_matched = line[len(key):].lstrip()
```
vvaltchev, line 49: If we use a regex for matching the keyword, that's just plain wrong. Just append ` (.*)` to the `keyword_expr` regex and rename it to `re_metadata_line`. It does not make sense to both use regex _and_ basic string manipulation funcs on the same string. Either one or the other. If you use the big cannon (regex), we could do also that with it.

vvaltchev (changes requested):
There are plenty of fixes of make, but overall the change is good.

scripts/configurator/parser/cmake_var.py
vvaltchev: Note: I added a comment below about this regex. It needs more changes than simply `^` at the beginning.

michelececcacci: Ok, i will refer to the comment below

scripts/configurator/parser/cmake_var.py
michelececcacci: Ok, but what about validation?

scripts/configurator/parser/cmake_var.py
michelececcacci: I used regex, it just makes sense here

scripts/configurator/parser/cmake_var.py
vvaltchev: What validation? That `var_name` really exists? You could do `assert(var_name in self.__dict__)`.
Note that `var_name` is not something that comes from the outer world. It's simply hard-coded in the dictionary:

```
"GROUP": self.get_setter("group")
```

Etc. So it's safe to do. Got it?

scripts/configurator/parser/cmake_var.py
```diff
@@ -1,9 +1,92 @@
...
+            msg = "{} is not a valid metadata keyword."
+            raise ValueError(msg.format(key))
+
+         elif self.kw_to_var[key] is not None:
+            msg = "Key {} has already a value associated to itself."
+            raise ValueError(msg.format(key))
+
+         setter = self.kw_to_setter[key]
+         s = match.group(2)
+         if key == "DEPENDS_ON":
```
vvaltchev, line 49: What's this? You should use the parser in the kw dict.
Like `p = kw[key].parser`  and later `setter(p(s))` or something like that.

scripts/configurator/parser/cmake_var.py
```diff
@@ -1,9 +1,92 @@
...
+      self.group: Optional[str] = None
+      self.realtype: Optional[str] = None
+
+      self.kw_to_setter: Mapping[str, Callable]= {
+         "REALTYPE": self.get_setter_realtype(),
+         "DEPENDS_ON": self.get_setter_depends_on(),
+         "GROUP": self.get_setter_group(),
+      }
+
+      self.kw_to_var: Mapping[str, Any] = {
```
vvaltchev, line 20: I don't understand why you dropped the `parser` dict. You should have replaced ALL the dicts with one and put the values in a namedTuple. Not just dropping the parser dict.

vvaltchev (changes requested):

scripts/configurator/parser/cmake_var.py
michelececcacci: What i meant was more like input validation. If i remember correctly, the user cannot pass any parameter to our setter and expect it to be correct

scripts/configurator/parser/cmake_var.py
michelececcacci: changed

scripts/configurator/parser/cmake_var.py
michelececcacci: changed

scripts/configurator/parser/cmake_var.py
michelececcacci: Ok. I suppose we will put this in another commit, right?

vvaltchev: Sure.

michelececcacci: Perfect, marking this as closed

scripts/configurator/parser/cmake_var.py
```diff
@@ -1,9 +1,84 @@
...
+               msg = """WARNING: First line uses a reserved keyword: {}.
+               This will be ignored."""
+               stderr.write(msg.format(key))
+            continue
+
+         elif key not in self.kw:
+            msg = "{} is not a valid metadata keyword."
+            raise ValueError(msg.format(key))
+
+         elif self.kw[key].var is not None:
```
vvaltchev, line 48: Do you have a test for this case? I'm not convinced that if the member variable changes in the object, the change will be reflected here, simply because python doesn't have explicit references: in kw we simply stored a reference to a particular value; after the member var is re-assigned, here we'll see the value of the old reference. That's why I suggested to have parser, setter and _getter_, where the getter is implemented similarly as the setter with `nonlocal` etc. 

You could simply test that by adding an `assert(self.kw[key].var is not None)` after calling the setter. I guess the assert will by triggered, because the var we have in `kw` will still be None.

vvaltchev (changes requested):
Much better, almost ready for merge. There is just this issue with the getter (see the comments). Please check that.

scripts/configurator/parser/cmake_var.py
michelececcacci: I will try, and correct if wrong

scripts/configurator/parser/cmake_var.py
```diff
@@ -1,9 +1,98 @@
...
+            raise ValueError(msg.format(key))
+
+         setter = self.kw[key].setter
+         parsed_value = self.kw[key].parser(match.group(2))
+         setter(parsed_value)
+         assert(self.kw[key].getter() is not None)
+
+   def get_setter(self, var_name: str):
+
+      def setter(val: List[str] | str) -> None:
```
vvaltchev: The type hints for `val` are not OK. This should really be `any`. Don't try to forcibly put a type even in places where the type could be _anything by design_. The point of type hints is to put a type where _there is_ type and it's always the same. When we want something to support any type, we shouldn't put the list of types that we currently support: that would be an anti-pattern. We're trying to take advantage, from time to time, of the dynamic nature of Python. In C we would have used a `void *` or a discriminated union because we had no better choice. In Java we would have just used `Object`.

scripts/configurator/parser/cmake_var.py
```diff
@@ -1,9 +1,98 @@
...
+         nonlocal self
+         self.__dict__[var_name] = val
+
+      return setter
+
+   def get_getter(self, var_name: str):
+
+      def getter():
+         nonlocal self
+
```
vvaltchev, line 73: Please remove this empty line.

vvaltchev (changes requested):
Almost perfect! Two more things to fix and I'll merge it.

scripts/configurator/parser/cmake_var.py
michelececcacci: Yeah, that's right. it's already removed. It also caused the CI to fail, because i was using the new syntax specific to 3.10

scripts/configurator/parser/cmake_var.py
michelececcacci: changed

vvaltchev (approved):
Perfect, I'm going to merge it now.

PR #133, opened 2022-08-28
[scripts] added type hinting to count_lines_of_code

michelececcacci:
While looking around I added some type hinting. Makes the code a bit more explicit in my opinion.

### DIFF
--- a/scripts/dev/count_lines_of_code
+++ b/scripts/dev/count_lines_of_code
@@ -4,6 +4,7 @@
 import os
 import sys
 import subprocess
+from typing import List, Optional, Dict, Tuple, Any
 
 # Constants
 
@@ -60,11 +61,11 @@ optPrintFileNames = False
 optShowLongLines = False
 optCheckLineEnding = False
 
-longLines = {} # file -> list of pairs (line#, cols)
+longLines: Dict[str, List[Tuple[int, int]]] = {} # file -> list of (line#, cols)
 nonUnixLeFiles = set()
 #####################################################
 
-def GetNoExtFileType(fpath):
+def GetNoExtFileType(fpath: str) -> Optional[str]:
 
    out = subprocess.check_output(['file', fpath]).decode('utf-8')
 
@@ -76,7 +77,7 @@ def GetNoExtFileType(fpath):
 
    return None
 
-def GetFileType(f, folder):
+def GetFileType(f: str, folder: str) -> Optional[str]:
 
    if f == "CMakeLists.txt":
       return "CMake"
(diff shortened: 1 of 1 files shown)


scripts/dev/count_lines_of_code
```diff
@@ -140,15 +141,15 @@ def IsDirExcluded(path):
...
    return fname in excludedFiles
 
-def AccumulateCounters(destArr, arr):
+def AccumulateCounters(destArr: Dict[str, int], arr: Dict[str, int]):
 
    for ft in fileTypes:
       destArr[ft] += arr[ft]
 
-def PrintRow(cols):
+def PrintRow(cols: List[str]) -> None:
```
vvaltchev, line 152: `PrintRow()` was supposed to take a `List` of `Any`, not a `List` of `str`, that's why it does `str(val)`. By forcing the type of the list, the `str(val)` because pointless. I'd make this `List[Any]`.

scripts/dev/count_lines_of_code
```diff
@@ -176,7 +177,7 @@ def PrintCounters(arr, eArr, total, eTot):
          PrintRow([ft] + vals + [GetEmptyPercStr(vals[0], vals[1])])
 
    print(("-" * 50))
-   PrintRow(["Total:", total, eTot, GetEmptyPercStr(total, eTot)])
+   PrintRow(["Total:", str(total), str(eTot), GetEmptyPercStr(total, eTot)])
```
vvaltchev, line 180: If you make `PrintRow` to accept as `List[Any]`, you won't need to do this extra conversions to string here.

scripts/dev/count_lines_of_code
```diff
@@ -161,10 +162,10 @@ def PrintRow(cols):
 
    print(s)
 
-def GetEmptyPercStr(t, e):
+def GetEmptyPercStr(t :int, e: int) -> str:
```
vvaltchev, line 165: **Formattin**g: this should be: `t: int`

vvaltchev (changes requested):
Thank you Michele, great job!

Note for changes like that: if you need to add explicit conversions like `str(val)` to make the linter pass, either the code is wrong or the new type hints are wrong. Try in most of the cases to add the type hints in a way that explicit conversions are not necessary. Of course I recognize that, sometimes, the right fix is actually to add the conversion.

scripts/dev/count_lines_of_code
michelececcacci: Ok, perfect

vvaltchev:
Thank you for the fixes!

vvaltchev (approved):

PR #136, opened 2022-09-10
[ci-fix] add gmock to toolchain

michelececcacci:
Opensuse, debian and fedora containers pass. The arch container needs to be regenerated, but this will be done with a dockerfile in another commit.

### DIFF
--- a/other/toolchain_builds.yml
+++ b/other/toolchain_builds.yml
@@ -33,6 +33,10 @@ steps:
     displayName: "[toolchain] Build gtest"
   - script: sudo -E ./scripts/build_toolchain -s build_gtest_arch
     displayName: "[toolchain] Build gtest_arch"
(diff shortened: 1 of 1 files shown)


vvaltchev (changes requested):
Good change, thank you for the fix.

Just, could you please reword the commit description? In the current form, the description is totally unclear to me:
 - `ci-fix` is a tag that has never been used before. You need a good reason to introduce a new tag. It's not just a random string.
 - `toolchain` doesn't mean anything. Actually, it might be confused with the build_toolchain script. In this case, you're touching toolchain_builds.yml, so be specific. 

An example is: 
```
[azure] Build gmock and gmock_arch in toolchain_builds.yml
```

I wonder if we should switch to the `[ci]` tag instead of `[azure]`.

michelececcacci:
Ok, i changed the commit's name to the one you suggested.

PR #137, opened 2022-09-10
[ci] added dockerfiles for containers in toolchain_build

michelececcacci:
Pretty self explanatory, all test pass (assuming CI has changes proposed in #136)

### DIFF
--- a/other/containers/arch-base/Dockerfile
+++ b/other/containers/arch-base/Dockerfile
@@ -1,0 +1,2 @@
+FROM archlinux:latest
+RUN pacman -Syu --noconfirm sudo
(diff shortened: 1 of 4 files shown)


vvaltchev (changes requested):
Great work, Michele!

A few changes are required though: 
 - in the change description, please don't use the name of the Azure pipeline, but the name of yml file, because that's in the repo. So, `toolchain_builds.yml`.
 - also, I suggest a rewording like: `[ci] add dockerfiles for toolchain_builds.yml containers`
 - I don't like the name of the containers: they are NOT tilck build containers: they are simply base images of debian, arch, fedora and opensuse with sudo. There is nothing tilck specific in them. Therefore, I'll call them debian-base, arch-base, etc. No tilck, nor build in their name.

michelececcacci:
used azure as commit tag, in order to be consistent with previous PR.

PR #138, opened 2022-09-10
[docker] Update the tilck-ci-build to v2

vvaltchev:
@michelececcacci

### DIFF
--- a/other/containers/tilck-ci-build/Dockerfile
+++ b/other/containers/tilck-ci-build/Dockerfile
@@ -1,17 +1,57 @@
-# version: v1
-# Also tried to use debian/stable-slim as base, but image size went up compared
-# to ubuntu 20.04
+# version: v2
 FROM ubuntu:20.04
+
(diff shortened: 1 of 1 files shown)


michelececcacci (approved):
LGTM. Personally, i would keep the comment  about how switching to debian slim actually increases image size(since it's unexpected). It's probably too verbose, but i think it's worth explaining. I actually find the comment on the debian frontend more unnecessary than the previously mentioned one.

vvaltchev:
> LGTM. Personally, i would keep the comment about how switching to debian slim actually increases image size(since it's unexpected). It's probably too verbose, but i think it's worth explaining. I actually find the comment on the debian frontend more unnecessary than the previously mentioned one.

As discussed offline, we cannot at the moment fully trust that comment we'll need to re-run a clean re-build of the image with debian-slim to check that. Therefore, in case of doubt, better remove the comment than having a misleading comment.

Issue #129, opened 2022-08-20
Add system tests for sys_setuid() and friends


vvaltchev:
As it can be seen here: https://app.codecov.io/gh/vvaltchev/tilck/blob/master/kernel/auth.c, the trivial functions in `kernel/auth.c` don't have any test coverage. That's because they are so simple that it's obvious why they work BUT, at the same time, it would still be nice to have test coverage for them for two reasons:
 1. It would increase Tilck's test coverage a little bit
 2. It would be a great exercise for new contributors to learn how to implement a new system test

To test `sys_getuid()`, `sys_setuid()` and friends, it would be enough to add new funcs in `tests/system/test_misc.c`. Actually, even a single `cmd_test_uids` could be enough to test them all.

Notes and hints:
 1. the new cmd function (shellcmd) must be registered in `tests/system/cmds_table.c`
 2. because simply calling `setuid()` from C would just call the latest syscall version, in order to test older syscalls like `sys_setgid16()`, it would be necessary to use `long syscall(long number, ...)` with the right syscall number from `kernel/arch/i386/arch_syscalls.c`.

PR #140, opened 2022-09-18
[gtests] add a test for getuid() & friends

michelececcacci:
References #129


tests/system/test_misc.c
```diff
@@ -322,3 +322,17 @@ int cmd_extra(int argc, char **argv)
 
    return rc;
 }
+
+int cmd_test_uids(void)
+{
+   int rc = 0;
+   rc |= syscall(SYS_getuid);
```
vvaltchev, line 329: This is not a competitive programming exercise :-)
Please use:
```
DEVSHELL_CMD_ASSERT(syscall(...) == 0);
```
It might look worse, but in reality is better because it will indicate exactly where the failure has happen. In tests, it matters a lot *how* we fail, not just that we either pass or fail.

vvaltchev (changes requested):
Thanks for that. Unfortunately, the x86-64 build fails:

```
/__w/1/s/tests/system/test_misc.c:332:18: error: 'SYS_getuid16' undeclared (first use in this function); did you mean 'SYS_getuid'?
    rc |= syscall(SYS_getuid16);
                  ^~~~~~~~~~~~
                  SYS_getuid
/__w/1/s/tests/system/test_misc.c:332:18: note: each undeclared identifier is reported only once for each function it appears in
[ 31%] Building C object kernel/arch/x86_64/CMakeFiles/acpica.dir/__/__/__/modules/acpi/acpica/executer/exconvrt.c.o
[ 31%] Building C object kernel/arch/x86_64/CMakeFiles/acpica.dir/__/__/__/modules/acpi/acpica/executer/excreate.c.o
[ 31%] Building C object kernel/arch/x86_64/CMakeFiles/acpica.dir/__/__/__/modules/acpi/acpica/executer/exdebug.c.o
[ 31%] Building C object kernel/arch/x86_64/CMakeFiles/acpica.dir/__/__/__/modules/acpi/acpica/executer/exdump.c.o
/__w/1/s/tests/system/test_misc.c:333:18: error: 'SYS_getgid16' undeclared (first use in this function); did you mean 'SYS_getgid'?
    rc |= syscall(SYS_getgid16);
                  ^~~~~~~~~~~~
                  SYS_getgid
/__w/1/s/tests/system/test_misc.c:334:18: error: 'SYS_geteuid16' undeclared (first use in this function); did you mean 'SYS_geteuid'?
    rc |= syscall(SYS_geteuid16);
                  ^~~~~~~~~~~~~
                  SYS_geteuid
/__w/1/s/tests/system/test_misc.c:335:18: error: 'SYS_getegid16' undeclared (first use in this function); did you mean 'SYS_getegid'?
    rc |= syscall(SYS_getegid16);
                  ^~~~~~~~~~~~~
                  SYS_getegid
```

Please, always check the CI logs in case something fails. Now that we have the ARCH=x86_64, it might fail too.

Btw, the reason is simple: the *16 syscalls does not exist for x86_64. Therefore, just #ifdef properly the calls referring them with `#ifdef __i386__`

tests/system/test_misc.c
michelececcacci: Perfect, i was actually wondering if there was a way to fail with more information. Turns out there is.

vvaltchev (changes requested):
Also, please change the commit message to something like: `[gtests] add a test for getuid() & friends`. The current description wasn't clear at all to me.

Finally, I don't like "test_uids" even if maybe I suggested it, in particular for the "test" prefix that is not used anywhere else in that list. Maybe just "getuids" is better?

vvaltchev (changes requested):
Great. It's almost ready to merge. Just, the commit description was supposed to be "[gtests] add a test for getuid() & friends" and not getuid**s**(). Please fix that.

Issue #135, opened 2022-09-09
Move other/musl-gcc and other/musl-g++ to scripts/templates


vvaltchev:
The `other/musl-gcc` and `other/musl-g++` template scripts are currently in the wrong place. The fix requires simply moving those two files to `scripts/templates` and replacing their paths in the `other/cmake/gen_config_pre.cmake` file:

Code snippet:

```cmake
smart_config_file(
   ${CMAKE_SOURCE_DIR}/other/musl-gcc
   ${MUSL_GCC}
)

smart_config_file(
   ${CMAKE_SOURCE_DIR}/other/musl-g++
   ${MUSL_GXX}
)
```

francescoparra:
Hello, may I take this issue?

vvaltchev:
Hello @francescoparra. Sure, absolutely :-)

vvaltchev:
Closing this issue, as it has been fixed by commit https://github.com/vvaltchev/tilck/commit/32b571d5625c04eefb9bf7a3ba732dc6c1e2cf53

PR #141, opened 2022-09-25
 [scripts] Move musl files to appropriate directory

francescoparra:
Related #135 

Moved musl files that were on a inappropriate directory to a more appropriate one and edited the paths in the `gen_config_pre.cmake` file


vvaltchev (changes requested):
Thanks for the change. The content itself is good but the head line of the commit message is way too long. It should be <= 74 chars. Also, we use tags like '[scripts]' here (I know, it's not mentioned in the contributing.md file).

So, could you make it be something like: 
`[scripts] Move musl-gcc and musl-g++ to scripts/templates/ (#135)` ?

Thank you

vvaltchev:
P.S. [optional] If you want to add more details, just leave the 2nd line as empty and add a commit description starting from the 3rd line. The head line is not supposed to tell everything about a commit ;-)

vvaltchev (changes requested):
Sorry man, but the commit message is still not OK. Let me explain why:

 - "musl files" is incorrect as the musl library is not part of this project. It is downloaded via a script in the toolchain. Those files are _template wrapper scripts_ for invoking gcc (and g++) with the toolchain-libmusl as default libc library (instead of glibc).
 - "appropriate directory" doesn't mean much, in particular here. (E.g. why `scripts/templates` is the appropriate directory instead of `other/`? Also, being very specific in commit messages helps a lot during bitsecting and history browsing; from this message it's not clear where those files has been moved and people would have to look at the actual diff to understand that.)
 - The new message is missing the issue number (`#135`), which is very useful to associate commits with issues fixed (see the edit below)

I suggested you a very good commit message:
`[scripts] Move musl-gcc and musl-g++ to scripts/templates/ (#135)`

It is concise and explains exactly what the change is doing[1]. For sure you could suggest an alternative message, but it has to be _better_ than mine. Doing so, would require a deep understanding of the context, which is not possible at the moment: it will take quite some time to understand the whole project etc. Therefore, just sticking with my suggestions is the fastest path for getting the pull request merged, at the beginning. Over time, if you like this project, things will get smoother and smoother. 

[1] the changes in the cmake file are a byproduct of the move operation.


**EDIT**: sorry, I didn't notice that you actually put the issue number in the commit message. My fault.

francescoparra:
Thanks and sorry for this incovenient problem, you could also make a specific section in the `contributing.md` on the commit message side so that you have a starting point for new contributors, or maybe there is another file containing this information that I haven't seen, in case let me know.

vvaltchev:
> Thanks and sorry for this incovenient problem, you could also make a specific section in the `contributing.md` on the commit message side so that you have a starting point for new contributors, or maybe there is another file containing this information that I haven't seen, in case let me know.

Thanks for the fix. Yeah, I'll definitively add a section in the `contributing.md` file. The problem is that it's very hard to explain in general terms (e.g. with rules) how to write a good commit message. I'll do my best mentioning the pre-existing tags, the idea of adding a reference to an issue (if any) etc., but that will not be exhaustive. It just takes time and experience to learn how to write a good commit message, in addition to domain knowledge on the specific project of course.

PR #143, opened 2022-10-10
[gtest] trigger the error code path in fork_dup_all_handles()

### LINKED ISSUE #120: Write a unit test to trigger the error code path in fork_dup_all_handles()
As it can be seen here: https://app.codecov.io/gh/vvaltchev/tilck/blob/master/kernel/fork.c, currently the following code path inside `fork_dup_all_handles()` is not covered by tests:

```C
      if (rc < 0 || !dup_h) {

         enable_preemption();
         {
            for (u32 j = 0; j < i; j++)
               vfs_close(pi->handles[j]);
         }
         disable_preemption();
         return -ENOMEM;
      }
```

Now that we have gMock in Tilck, it would be relatively simple to just mock `vfs_dup` and `vfs_close()` so that inside a unit test we can trigger this code path. To do that, it would be necessary to make the `fork_dup_all_handles()` function `STATIC` instead of `static`, so that in unit tests, the function won't be static anymore. Also, it would be necessary add the mocked functions to a CMakeLists.txt file and to another file in the unit tests directory. For the exact instructions, see the comments in: https://github.com/vvaltchev/tilck/blob/master/tests/unit/mock_experiments.cpp

Clearly, because the `fork_dup_all_handles()` function takes a `struct process *` parameter, the unit test will have to allocate such a structure and set at least 2 handles in the `pi->handles` array.

michelececcacci:
Closes #120 . [Hits the path as required](https://app.codecov.io/gh/michelececcacci/tilck/blob/gmock/kernel/fork.c) and no memory leaks. Also adding CI run [link](https://vkvaltchev.visualstudio.com/Tilck/_build/results?buildId=1766&view=logs&j=dbef4360-82de-57cb-1dd1-01eea5f538b1&t=38631106-5089-5c4e-5921-7d2ecd523aa3)


vvaltchev (changes requested):
Good work! A great improvement compared to the past. The change still requires some work, but it's close-enough to its final shape :-)

tests/unit/fork_test.cpp
```diff
@@ -0,0 +1,40 @@
+#include <gtest/gtest.h>
```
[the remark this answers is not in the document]
michelececcacci: Added

tests/unit/fork_test.cpp
```diff
@@ -0,0 +1,40 @@
...
+#include <gmock/gmock.h>
+
+using namespace testing;
+
+extern "C" {
+   #include <unistd.h>
```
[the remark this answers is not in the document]
michelececcacci: I moved `unistd.h` above gtest and gmock.

tests/unit/fork_test.cpp
```diff
@@ -0,0 +1,40 @@
...
+
+   EXPECT_CALL(mock, vfs_dup(_, _))
+      .WillOnce(DoAll(
+               SetArgPointee<1>(allocated_h),
+               Return(0)
+               ))
+      .WillOnce(Return(-1));
+   EXPECT_CALL(mock, vfs_close(_));
+   ASSERT_EQ(fork_dup_all_handles(pi), -ENOMEM);
+   free(allocated_h);
```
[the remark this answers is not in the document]
michelececcacci: Done, switched to stack allocatoins

tests/unit/fork_test.cpp
```diff
@@ -0,0 +1,40 @@
...
+   pi->handles[0] = malloc(sizeof(fs_handle));
+   pi->handles[1] = malloc(sizeof(fs_handle));
+
+   EXPECT_CALL(mock, vfs_dup(_, _))
+      .WillOnce(DoAll(
+               SetArgPointee<1>(allocated_h),
+               Return(0)
+               ))
+      .WillOnce(Return(-1));
+   EXPECT_CALL(mock, vfs_close(_));
```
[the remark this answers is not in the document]
michelececcacci: I was wondering? should we also make assumptions on the arguments of `vfs_dup`? We could use `RetiresOnSaturation()`, but that may be too strict.

tests/unit/fork_test.cpp
```diff
@@ -0,0 +1,52 @@
...
+
+TEST(fork_dup_all_handles, trigger_inside_path) {
+   vfs_mock mock;
+   process pi;
+   bzero(&pi, sizeof(process));
+   fs_handle_base handle0_dup, handle1_dup, handle2_dup;
+
+   pi.handles[0] = (fs_handle) &handle0_dup;
+   pi.handles[1] = (fs_handle) &handle1_dup;
+   pi.handles[2] = (fs_handle) &handle2_dup;
```
vvaltchev, line 32: **BUG:** Actually, that's wrong here. Here we should set the original handles, not the duplicated ones. We need theoretically 3 handles (original) and 2 duplicate ones. You can fake that this way:

```
   pi.handles[0] = TO_PTR(0xaa00);
   pi.handles[1] = TO_PTR(0xaa01);
   pi.handles[2] = TO_PTR(0xaa02);
```

Those pointers are **invalid**, but that's fine because we mocked `vfs_dup()`.

tests/unit/fork_test.cpp
```diff
@@ -0,0 +1,52 @@
+/* SPDX-License-Identifier: BSD-2-Clause */
+
+#include <unistd.h>
```
vvaltchev, line 3: Better, but just out of curiosity: why do we need unistd.h in this specific file?

tests/unit/fork_test.cpp
```diff
@@ -0,0 +1,52 @@
...
+      )
+      .WillOnce(
+         DoAll(
+               SetArgPointee<1>(&handle1_dup),
+               Return(0)
+         )
+      )
+      .WillOnce(Return(-1));
+
+   EXPECT_CALL(mock, vfs_close(&handle0_dup));
```
vvaltchev, line 49: Theoretically, we could totally fake the dup handles and make `vfs_dup()` to return 0xbb00, 0xbb01, 0xbb02. Not sure what's best, really. Maybe allocate 3 actual handles + 2 dup handles, just to stay closer to reality, even if those pointers are never de-referenced.

tests/unit/misc_fake.c
```diff
@@ -88,7 +89,7 @@ void __wrap_kmutex_unlock(struct kmutex *m) {
 }
 
 /*
- * Decide with just a global flag wheather to use glibc's malloc() or Tilck's
+ * Decide with just a global flag whether to use glibc's malloc() or Tilck's
```
vvaltchev, line 92: Thanks!!

tests/unit/fork_test.cpp
vvaltchev: Or, as suggested below, if you prefer the test to be more realistic and less "tricky", you can just allocate an array of 3 handles and another array of 2 dup_handles.

tests/unit/fork_test.cpp
```diff
@@ -0,0 +1,52 @@
...
+}
+
+class vfs_mock : public KernelSingleton {
+public:
+
+   MOCK_METHOD(int, vfs_dup, (fs_handle h, fs_handle *dup_h), (override));
+   MOCK_METHOD(void, vfs_close, (fs_handle h), (override));
+};
+
+TEST(fork_dup_all_handles, trigger_inside_path) {
```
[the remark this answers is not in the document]
michelececcacci: Done

tests/unit/fork_test.cpp
```diff
@@ -0,0 +1,52 @@
...
+   fs_handle_base handle0_dup, handle1_dup, handle2_dup;
+
+   pi.handles[0] = (fs_handle) &handle0_dup;
+   pi.handles[1] = (fs_handle) &handle1_dup;
+   pi.handles[2] = (fs_handle) &handle2_dup;
+
+   EXPECT_CALL(mock, vfs_dup(_, _))
+      .WillOnce(
+         DoAll(
+               SetArgPointee<1>(&handle0_dup),
```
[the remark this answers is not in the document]
michelececcacci: I didn't copy paste your code, but i tried to replicate indentation. Fixed. I assume it should also be the case for Return, right?

tests/unit/fork_test.cpp
```diff
@@ -0,0 +1,52 @@
...
+
+#include <unistd.h>
+#include <gtest/gtest.h>
+#include <gmock/gmock.h>
+
+using namespace testing;
+
+extern "C" {
+   #include <tilck/kernel/process.h>
+   #include "mocking.h"
```
vvaltchev, line 11: **FIXIT** mocking.h is a C++ header, designed to be used for the C++ tests, so it shouldn't be wrapped by extern "C". Put it after `#include <gmock/gmock.h>`

michelececcacci: Oh, ok. My bad

tests/unit/fork_test.cpp
```diff
@@ -0,0 +1,52 @@
...
+
+#include <gtest/gtest.h>
+#include <gmock/gmock.h>
+#include "mocking.h"
+
+using namespace testing;
+
+extern "C" {
+   #include <tilck/kernel/process.h>
+   int vfs_dup(fs_handle h, fs_handle *dup_h);
```
vvaltchev, line 11: Not 100% sure, but I guess that `vfs_dup()` and `vfs_close()` are not required to be defined there. As mocked methods, they should be defined by `mocking.h`. Could you please check? If they are not required, just remove those definitions from here. Otherwise, let me know.

tests/unit/misc_fake.c
```diff
@@ -9,6 +9,7 @@
 #include <tilck/kernel/datetime.h>
 #include <tilck/kernel/sync.h>
 #include <tilck/kernel/sched.h>
+#include <tilck/kernel/fs/vfs_base.h>
```
vvaltchev, line 12: **FIXME:** Why including `vfs_base.h` here is necessary? There are no new changes in misc_fake.c that require it.

vvaltchev (changes requested):
Great job, Michele! The change is almost ready to be shipped :-)
There are just two minor things to address.

vvaltchev (approved):
Perfect! I'm merging this.

Issue #145, opened 2022-10-12
Add unit tests for functions in string_util.c


vvaltchev:
As it can be seen here: https://app.codecov.io/gh/vvaltchev/tilck/blob/master/common/string_util.c, those functions do not have full coverage, while it will good to have that.

One of the technical problems here is that those functions have the same name as the standard libc functions (e.g. strcpy(), strcat() etc.). Therefore, they cannot be exported to unit tests directly. The solution to this problem is simple:
 - add a `tilck_` prefix to those functions
 - under `#if !defined(TESTING) && !defined(USERMODE_APP)`, create alias symbols using the GCC alias attribute.

Example:
```C
int tilck_add(int a, int b)
{
	return a+b;
}

#if !defined(TESTING) && !defined(USERMODE_APP)
int add(int a, int b) __attribute__((alias("tilck_add"))); 
#endif
```

This way, it will be trivial to test those functions from C++, using they full name (tilck_strcmp, tilck_strcat, etc.). The goal is having 100% line coverage.

michelececcacci:
This should be closed as completed

vvaltchev:
Sure.

PR #148, opened 2022-10-29
[gtest] added tests for string_util (#145)

michelececcacci:
100% coverage as required. Maybe variable naming should be improved, but i think that' s about it.
https://app.codecov.io/gh/michelececcacci/tilck/blob/addtest/common/string_util.c

### DIFF
--- a/common/string_util.c
+++ b/common/string_util.c
@@ -79,7 +79,7 @@ int memcmp(const void *_m1, const void *_m2, size_t n)
    return i == n ? 0 : (int)*m1 - (int)*m2;
 }
 
-char *strstr(const char *haystack, const char *needle)
+char *tilck_strstr(const char *haystack, const char *needle)
 {
    size_t sl, nl;
 
@@ -101,6 +101,12 @@ char *strstr(const char *haystack, const char *needle)
    return NULL;
 }
 
+#if !defined(TESTING) && !defined(USERMODE_APP)
+   char *strstr(const char *haystack, const char *needle) \
+   __attribute__((alias("tilck_strstr")));
+#endif
+
+
 char *strcpy(char *dest, const char *src)
 {
    char *p = dest;
@@ -112,7 +118,7 @@ char *strcpy(char *dest, const char *src)
    return dest;
 }
 
-char *strncpy(char *dest, const char *src, size_t n)
+char *tilck_strncpy(char *dest, const char *src, size_t n)
 {
    char *p = dest;
    size_t i = 0;
@@ -128,12 +134,17 @@ char *strncpy(char *dest, const char *src, size_t n)
    return dest;
 }
 
+#if !defined(TESTING) && !defined(USERMODE_APP)
+   char *strncpy(char *dest, const char *src, size_t n) \
+   __attribute__((alias("tilck_strncpy")));
+#endif
+
 char *strcat(char *dest, const char *src)
 {
    return strcpy(dest + strlen(dest), src);
 }
 
-char *strncat(char *dest, const char *src, size_t n)
+char *tilck_strncat(char *dest, const char *src, size_t n)
 {
    char *p = dest + strlen(dest);
    size_t i = 0;
@@ -147,17 +158,29 @@ char *strncat(char *dest, const char *src, size_t n)
    return dest;
 }
 
+#if !defined(TESTING) && !defined(USERMODE_APP)
+   char *strncat(char *dest, const char *src, size_t n) \
+   __attribute__((alias("tilck_strncat")));
+#endif
 
-int isxdigit(int c)
+int tilck_isxdigit(int c)
 {
    return c < 128 && digit_to_val[c] >= 0;
 }
 
-int isspace(int c)
+#if !defined(TESTING) && !defined(USERMODE_APP)
+   int isxdigit(int c) __attribute__((alias("tilck_isxdigit")));
+#endif
+
+int tilck_isspace(int c)
 {
    return c == ' ' || c == '\t' || c == '\r' ||
           c == '\n' || c == '\v' || c == '\f';
 }
+
+#if !defined(TESTING) && !defined(USERMODE_APP)
+   int isspace(int c) __attribute__((alias("tilck_isspace")));
+#endif
 
 #endif // #if !defined(TESTING) && !defined(USERMODE_APP)
 
--- a/include/tilck/common/string_util.h
+++ b/include/tilck/common/string_util.h
@@ -113,3 +113,13 @@ void uitoaN_hex_fixed(ulong value, char *buf);    /* pointer-size */
 
 long tilck_strtol(const char *str, const char **endptr, int base, int *error);
 ulong tilck_strtoul(const char *str, const char **endptr, int base, int *error);
+int tilck_strcmp(const char *s1, const char *s2);
+int tilck_strncmp(const char *s1, const char *s2, size_t n);
+int tilck_memcmp(const void *m1, const void *m2, size_t n);
+char *tilck_strstr(const char *haystack, const char *needle);
+char *tilck_strcpy(char *dest, const char *src);
+char *tilck_strncpy(char *dest, const char *src, size_t n);
+char *tilck_strcat(char *dest, const char *src);
+char *tilck_strncat(char *dest, const char *src, size_t n);
+int tilck_isxdigit(int c);
+int tilck_isspace(int c);
--- a/tests/unit/string_util_test.cpp
+++ b/tests/unit/string_util_test.cpp
@@ -1,0 +1,57 @@
+/* SPDX-License-Identifier: BSD-2-Clause */
+
+#include <gtest/gtest.h>
+
+using namespace testing;
+
+extern "C" {
+   #include <tilck/common/string_util.h>
+}
+
+TEST(strstr, basic)
+{
+   const char haystack[] = "haystack", needle[] = "needle", \
+   hay[] = "hay", empty[] = "";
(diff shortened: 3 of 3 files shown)


tests/unit/string_util_test.cpp
```diff
@@ -0,0 +1,57 @@
...
+
+extern "C" {
+   #include <tilck/common/string_util.h>
+}
+
+TEST(strstr, basic)
+{
+   const char haystack[] = "haystack", needle[] = "needle", \
+   hay[] = "hay", empty[] = "";
+   ASSERT_EQ(strstr(empty, needle), (char *) NULL);
```
vvaltchev, line 15: You should really call tilck_strstr() here.

tests/unit/string_util_test.cpp
```diff
@@ -0,0 +1,57 @@
...
+   ASSERT_EQ(strstr(haystack, needle), (char *) NULL);
+   ASSERT_EQ(strstr(haystack, hay), haystack);
+}
+
+TEST(strncpy, basic)
+{
+   char dest[4], src[] = "a";
+   strncpy(dest, src,2);
```
vvaltchev, line 23: You’re calling libc’s strncpy() implementation. You should really rename this symbol to tilck_strncpy() and add an alias as you did for strstr().

tests/unit/string_util_test.cpp
```diff
@@ -0,0 +1,57 @@
...
+   ASSERT_EQ(0, strcmp(dest, "a"));
+}
+
+TEST(strncat, basic)
+{
+   char dest[6] = "abc", src[] = "d";
+   strncat(dest, src, 4);
+   ASSERT_EQ(0, strcmp(dest, "abcd"));
```
vvaltchev, line 31: I believe there is something like ASSERT_STREQ, which is way more convenient than this.

tests/unit/string_util_test.cpp
```diff
@@ -0,0 +1,57 @@
...
+}
+
+TEST(isxdigit, basic)
+{
+   ASSERT_EQ(isxdigit(48), true);
```
vvaltchev, line 36: Same as above. How can we be absolutely sure you’re calling Tilck’s implementation of this standard function? Even if in this specific contex for a series of reasons you might be actually calling Tilck code, this approach is fragile. Prefix all of those functions and add aliases.

tests/unit/string_util_test.cpp
```diff
@@ -0,0 +1,57 @@
...
+   ASSERT_EQ(isxdigit(58), false);
+   ASSERT_EQ(isxdigit(71), false);
+   ASSERT_EQ(isxdigit(127), false);
+   ASSERT_EQ(isxdigit(128), false);
+}
+
+TEST(isspace, basic)
+{
```
vvaltchev, line 44: Same.

common/string_util.c
```diff
@@ -79,7 +79,7 @@ int memcmp(const void *_m1, const void *_m2, size_t n)
    return i == n ? 0 : (int)*m1 - (int)*m2;
 }
 
-char *strstr(const char *haystack, const char *needle)
+char *tilck_strstr(const char *haystack, const char *needle)
```
vvaltchev, line 82: Good job here.

vvaltchev (changes requested):

common/string_util.c
```diff
@@ -101,6 +101,12 @@ char *strstr(const char *haystack, const char *needle)
    return NULL;
 }
 
+#if !defined(TESTING) && !defined(USERMODE_APP)
+   char *strstr(const char *haystack, const char *needle) \
```
vvaltchev, line 105: Good job here. Just, I don’t understand why you did that only for strstr() and not for the other functions. See the other comments.

vvaltchev:
Final comment: note that you’ll need to add definitions of the tilck_* functions in the string util header file (don’t remember its name now, I’m using the phone) plus you’ll need to add some function macros wrappers that are visibile only within the kernel. 

Not sure what to do with the inline functions when there is name clashing. Think about that. Either with macros or aliases we should be able to handle that case too.

tests/unit/string_util_test.cpp
michelececcacci: You're right. Fixed this.

michelececcacci:
> Final comment: note that you’ll need to add definitions of the tilck_* functions in the string util header file (don’t remember its name now, I’m using the phone) plus you’ll need to add some function macros wrappers that are visibile only within the kernel.

Not really sure why, but it looks like they just need to be included as extern.

tests/unit/string_util_test.cpp
```diff
@@ -0,0 +1,62 @@
+/* SPDX-License-Identifier: BSD-2-Clause */
+
+#include <gtest/gtest.h>
+
+using namespace testing;
+
+extern "C" {
+   #include <tilck/common/string_util.h>
+   char *tilck_strstr(const char *haystack, const char *needle);
```
vvaltchev, line 9: Of course declaring here the `tilck_*` functions works, but it's not safe. That's why I said that you have to put those in the string_util.h header.

Why it's not safe? Because if for any reason somebody changes the signature of one of those functions in the kernel without fixing these ones, no warnings or errors will be shown. That's simply because these declarations are not visible while compiling their bodies. It's like declaring a function twice in two separate C files, instead of using a header. **It's very dangerous.** [1]

Therefore, add those in `string_util.h` and add the short names (e.g. strcmp etc.) under `#if !defined(TESTING) && !defined(USERMODE_APP)`. At the beginning of that header file there is already such a block. Just append them at the bottom of it and add the `tilck_*` ones outside of the `#ifdef`.  Makes sense?

----------------------
[1] It's true that sometimes I might have told you to just declare the kernel function in the unit test. That applies for STATIC functions that are not declared in any header. And it's still dangerous for the same reasons. I'm starting to think that we should introduce special test-only headers with all of those declarations, just to get that extra compile-time safety.

vvaltchev (changes requested):

tests/unit/string_util_test.cpp
michelececcacci: Sounds good. Sorry for the extra hassle, but i want to make sure i understand the reasoning behind some choices.

include/tilck/common/string_util.h
```diff
@@ -29,14 +29,28 @@
...
 
-int strcmp(const char *s1, const char *s2);
-int strncmp(const char *s1, const char *s2, size_t n);
-int memcmp(const void *m1, const void *m2, size_t n);
-char *strstr(const char *haystack, const char *needle);
-char *strcpy(char *dest, const char *src);
-char *strncpy(char *dest, const char *src, size_t n);
-char *strcat(char *dest, const char *src);
-char *strncat(char *dest, const char *src, size_t n);
+#if !defined(TESTING) && !defined(USERMODE_APP)
```
vvaltchev, line 32: You're already inside of a `#if !defined(TESTING) && !defined(USERMODE_APP)` block.
It doesn't make any sense to add another such `#if` here. Please, check better the context on your changes.

include/tilck/common/string_util.h
```diff
@@ -29,14 +29,28 @@
...
+   char *strcpy(char *dest, const char *src);
+   char *strncpy(char *dest, const char *src, size_t n);
+   char *strcat(char *dest, const char *src);
+   char *strncat(char *dest, const char *src, size_t n);
+
+   int isxdigit(int c);
+   int isspace(int c);
+#endif
+
+int tilck_strcmp(const char *s1, const char *s2);
```
vvaltchev, line 46: These were supposed to be **outside** of the `#if !defined(TESTING) && !defined(USERMODE_APP)` block, right? It ends a few lines below. There is also a comment.

tests/unit/string_util_test.cpp
```diff
@@ -0,0 +1,62 @@
...
+   char *tilck_strncpy(char *dest, const char *src, size_t n);
```
vvaltchev, line 10: The whole point of putting those `tilck_*` declarations inside the header is to remove them from this file, because it's unsafe to put them here (there are no compile-time checks validating that the signature matches).

vvaltchev (changes requested):

include/tilck/common/string_util.h
michelececcacci: Ok, sorry

tests/unit/string_util_test.cpp
michelececcacci: sorry, again.

include/tilck/common/string_util.h
michelececcacci: Btw, i think always indenting when inside of an ifdef can prevent this problem in the future. I think you considered that already, but i just wanted to point it out

tests/unit/string_util_test.cpp
```diff
@@ -0,0 +1,57 @@
...
+
+using namespace testing;
+
+extern "C" {
+   #include <tilck/common/string_util.h>
+}
+
+TEST(strstr, basic)
+{
+   const char haystack[] = "haystack", needle[] = "needle", \
```
vvaltchev, line 13: I forgot to mention this cosmetic detail, but this style is really UGLY to me. Don't need to fix it as I'm preparing a change now and will fix it too. Just, for the next time, do something like:

```C
const char *haystack = "haystack";
const char *needle = "needle";
const char *hay = "hay";
const char *empty = "";
```

Better to split the declarations across multiple lines and try NEVER to use the backslash. In this case it was also **totally pointless**.

michelececcacci: Ok, perfect. Sorry for the inconvenience

PR #150, opened 2022-11-27
[gtest] add tests for tracing functions (#146)

### LINKED ISSUE #146: Add tests for simple_wildcard_match() and set_traced_syscalls_int()
As it can be seen here: https://app.codecov.io/gh/vvaltchev/tilck/blob/master/modules/tracing/tracing.c, the functions `simple_wildcard_match()` and `set_traced_syscalls_int()` have little test coverage, while it will be easy to write unit tests for them.

Just make the functions be `STATIC` instead of `static`, so that the functions will be visible in the unit tests binary. The goal here is to reach 100% line coverage for both the functions. 

HINT: testing  `set_traced_syscalls_int()` will require mocking `handle_sys_trace_arg()`, which is called by it.

michelececcacci:
Closes #146. Codecov report [here](https://app.codecov.io/gh/michelececcacci/tilck/blob/addtest/modules/tracing/tracing.c/). Reached 100% coverage just with unit tests, but we may need more extensive testing.

### DIFF
--- a/include/tilck/kernel/test/tracing.h
+++ b/include/tilck/kernel/test/tracing.h
@@ -1,0 +1,10 @@
+/* SPDX-License-Identifier: BSD-2-Clause */
+
+#pragma once
+
+#include <tilck/mods/tracing.h>
+STATIC bool simple_wildcard_match(const char *str, const char *expr);
+STATIC int set_traced_syscalls_int(const char *str);
+
+extern bool *traced_syscalls;
+extern char *traced_syscalls_str;
--- a/tests/unit/tracing_test.cpp
+++ b/tests/unit/tracing_test.cpp
@@ -1,0 +1,64 @@
+/* SPDX-License-Identifier: BSD-2-Clause */
+
+#include <gtest/gtest.h>
+#include <gmock/gmock.h>
+#include "mocking.h"
+
+using namespace testing;
+
+extern "C" {
+   #include <tilck/kernel/test/tracing.h>
+}
+
+
+class tracing_mock : public KernelSingleton {
+public:
+   MOCK_METHOD(int, handle_sys_trace_arg, (const char *arg), (override));
+};
+
+TEST(simple_wildcard_match, basic)
+{
+   ASSERT_TRUE(simple_wildcard_match("test string", "test string"));
+   ASSERT_TRUE(simple_wildcard_match("test string", "test string*"));
+   ASSERT_TRUE(simple_wildcard_match("test string", "tes? string"));
+   ASSERT_TRUE(simple_wildcard_match("test string", "test *"));
+   ASSERT_FALSE(simple_wildcard_match("test string", "test strin1"));
+   ASSERT_FALSE(simple_wildcard_match("test string", "te*t"));
+   ASSERT_FALSE(simple_wildcard_match("test string", "test string1"));
+}
+
+TEST(set_traced_syscalls_int, basic)
+{
+   tracing_mock mock;
+   char tracing_str[TRACED_SYSCALLS_STR_LEN+1];
+   bool traced_syscalls_mock[MAX_SYSCALLS] = {1};
+
+   memset((void *) tracing_str, 'a', TRACED_SYSCALLS_STR_LEN);
+   tracing_str[TRACED_SYSCALLS_STR_LEN] = '\0';
+   traced_syscalls = traced_syscalls_mock;
+   traced_syscalls_str = tracing_str;
+
+   ASSERT_EQ(set_traced_syscalls_int(tracing_str), -ENAMETOOLONG);
+
+   tracing_str[TRACED_SYSCALLS_STR_LEN-1] = '\0';
+   ASSERT_EQ(set_traced_syscalls_int(tracing_str), -ENAMETOOLONG);
+
+   EXPECT_CALL(mock, handle_sys_trace_arg(_))
+      .WillOnce(Return(1));
+   ASSERT_EQ(set_traced_syscalls_int(",,,"), 1);
+
+   EXPECT_CALL(mock, handle_sys_trace_arg(_))
+      .WillOnce(Return(1));
+   ASSERT_EQ(set_traced_syscalls_int("abc"), 1);
+
+   EXPECT_CALL(mock, handle_sys_trace_arg(_))
+      .WillOnce(Return(0));
+   ASSERT_EQ(set_traced_syscalls_int("abc"), 0);
+
+   EXPECT_CALL(mock, handle_sys_trace_arg(_))
+      .WillRepeatedly(Return(0));
+   ASSERT_EQ(set_traced_syscalls_int("a, b, c"), 0);
+
+   traced_syscalls = NULL;
+   traced_syscalls_str = NULL;
+}
--- a/modules/tracing/tracing.c
+++ b/modules/tracing/tracing.c
@@ -40,7 +40,7 @@ static const struct syscall_info **syscalls_info;
 static s8 (*params_slots)[MAX_SYSCALLS][6];
 static s8 *syscalls_fmts;
 
-static char *traced_syscalls_str;
+STATIC char *traced_syscalls_str;
 static int traced_syscalls_count;
 
 bool *traced_syscalls;
@@ -634,7 +634,7 @@ get_traced_syscalls_count(void)
  *    ab*c     matches NOTHING because '*' must be the last char, if present.
  *
  */
-static bool
+STATIC bool
 simple_wildcard_match(const char *str, const char *expr)
 {
    for (; *str && *expr; str++, expr++) {
@@ -657,7 +657,7 @@ simple_wildcard_match(const char *str, const char *expr)
    return false; /* not a match */
 }
 
-static int
+STATIC int
 handle_sys_trace_arg(const char *arg)
 {
    struct bintree_walk_ctx ctx;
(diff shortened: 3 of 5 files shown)


tests/unit/tracing_test.cpp
```diff
@@ -0,0 +1,74 @@
...
+#include <gmock/gmock.h>
+#include "mocking.h"
+
+using namespace testing;
+
+extern "C" {
+   #include <tilck/kernel/test/tracing.h>
+}
+
+extern bool *traced_syscalls;
```
vvaltchev, line 13: The whole point of introducing `include/tilck/kernel/test/tracing.h` and all the files there is to avoid unchecked declartions like those in the unit tests. Move the `extern bool *traced_syscalls;` to that file as it is. While for `traced_syscalls_str` you need to declare it there with STATIC.

tests/unit/tracing_test.cpp
```diff
@@ -0,0 +1,74 @@
...
+extern char *traced_syscalls_str;
+
+class tracing_mock : public KernelSingleton {
+public:
+   MOCK_METHOD(int, handle_sys_trace_arg, (const char *arg), (override));
+};
+
+TEST(simple_wildcard_match, basic)
+{
+   const char *str1 = "test string";
```
vvaltchev, line 23: Don't name test strings like that, please. It's hard to read for me. Put all the test strings as literal while calling `simple_wildcard_match`. Also, use more than one string to match against.

tests/unit/tracing_test.cpp
```diff
@@ -0,0 +1,74 @@
...
+   const char *str1_equivalent = "test string*";
+   const char *qmark_str1 = "tes? string";
+   const char *wrong_str1 = "test string1";
+   const char *asterisk_str1 = "test *";
+   const char *str2 = "test strin1";
+   const char *invalid_str = "te*t ";
+
+   ASSERT_EQ(true, simple_wildcard_match(str1, str1));
```
vvaltchev, line 31: Please, use ASSERT_TRUE(). Also, as you can see yourself, you're testing only for positive cases. That's not enough.

tests/unit/tracing_test.cpp
```diff
@@ -0,0 +1,74 @@
...
+   ASSERT_EQ(true, simple_wildcard_match(str1, asterisk_str1));
+   ASSERT_EQ(false, simple_wildcard_match(str1, str2));
+   ASSERT_EQ(false, simple_wildcard_match(str1, invalid_str));
+   ASSERT_EQ(false, simple_wildcard_match(str1, wrong_str1));
+}
+
+TEST(set_traced_syscalls_int, basic)
+{
+   tracing_mock mock;
+   const char *too_long = "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa"
```
vvaltchev, line 43: That's very bad.
1. Find where `TRACED_SYSCALLS_STR_LEN` and include that header here
2. Create an array on-stack like `const char too_long[TRACED_SYSCALLS_STR_LEN + 1];` and initialize it with memset.
3. Do the same for `almost_too_long`.

tests/unit/tracing_test.cpp
```diff
@@ -0,0 +1,74 @@
...
+
+   EXPECT_CALL(mock, handle_sys_trace_arg(_))
+      .WillOnce(Return(1));
+   ASSERT_EQ(set_traced_syscalls_int(str1), 1);
+
+   EXPECT_CALL(mock, handle_sys_trace_arg(_))
+      .WillOnce(Return(1));
+   ASSERT_EQ(set_traced_syscalls_int(normal_name), 1);
+
+   traced_syscalls = (bool *) malloc(sizeof(bool) * MAX_SYSCALLS); // TODO clean up memory
```
vvaltchev, line 62: Nope.
1. It's safe to assume that sizeof(bool) == 1
2. Don't use the heap, unless there's no other choice. You can make that point to a stack array on the stack and set to NULL at the end of this function.

tests/unit/tracing_test.cpp
```diff
@@ -0,0 +1,74 @@
...
+   bzero(traced_syscalls, sizeof(bool) * MAX_SYSCALLS);
+   traced_syscalls[0] = 1;
+   traced_syscalls_str = (char *) malloc(1);
```
vvaltchev, line 65: OMG no! `malloc(1)` is crime. What is that?

vvaltchev (changes requested):
There are many things to improve in this change. The first of all is recognizing that it broke two targets: https://vkvaltchev.visualstudio.com/Tilck/_build/results?buildId=1814&view=logs&jobId=92da376a-0515-5e27-5d1e-1ffcb61f64c8&j=92da376a-0515-5e27-5d1e-1ffcb61f64c8&t=fa1dd8db-7c27-510b-5998-0b4baae709d9

tests/unit/tracing_test.cpp
michelececcacci: Fixed.

tests/unit/tracing_test.cpp
michelececcacci: Fixed

tests/unit/tracing_test.cpp
michelececcacci: So should i just insert directly in the function calls?

vvaltchev: Yes, absolutely. It will be way more clear that way.

tests/unit/tracing_test.cpp
michelececcacci: I used ASSERT_TRUE and ASSERT_FALSE. But i will probably write more test cases (doesn't hurt)

tests/unit/tracing_test.cpp
michelececcacci: The idea is not to segfault when using memset (needed to get 100% coverage). I could also just allocate a single character and pass its address. That's probably better

tests/unit/tracing_test.cpp
```diff
@@ -0,0 +1,61 @@
...
+   ASSERT_TRUE(simple_wildcard_match("test string", "test *"));
+   ASSERT_FALSE(simple_wildcard_match("test string", "test strin1"));
+   ASSERT_FALSE(simple_wildcard_match("test string", "te*t"));
+   ASSERT_FALSE(simple_wildcard_match("test string", "test string1"));
+}
+
+TEST(set_traced_syscalls_int, basic)
+{
+   tracing_mock mock;
+   const char too_long[TRACED_SYSCALLS_STR_LEN] = {'a'};
```
vvaltchev, line 33: 1. Don't initialize this to with `{'a'}` as it's doesn't make any sense. You're using memset already.
2. You are not zero-terminating the string after memset.
3. You don't really need two arrays like that. It's enough to have just one and calling it `tracing_str` and then make it shorter by moving the zero-termination.
4. Please, learn to use vertical space. 
5. Separate variable declarations from memset calls. First the array and the other variables, then vertical space, then the memset calls and the other stuff.

tests/unit/tracing_test.cpp
```diff
@@ -0,0 +1,61 @@
...
+
+   EXPECT_CALL(mock, handle_sys_trace_arg(_))
+      .WillOnce(Return(1));
+   ASSERT_EQ(set_traced_syscalls_int(",,,"), 1);
+
+   EXPECT_CALL(mock, handle_sys_trace_arg(_))
+      .WillOnce(Return(1));
+   ASSERT_EQ(set_traced_syscalls_int("abc"), 1);
+
+   bool traced_syscalls_mock[MAX_SYSCALLS] = {0};
```
vvaltchev, line 48: Put this one the top with the other stack variables.

tests/unit/tracing_test.cpp
```diff
@@ -0,0 +1,61 @@
...
+   traced_syscalls = traced_syscalls_mock;
+   traced_syscalls[0] = 1;
+   char single_char = 'a';
+   traced_syscalls_str = &single_char;
```
vvaltchev, line 52: This still doesn't make any sense. I had to check the code and try to figure out myself what you're trying to achieve. "Not crashing" is not enough. The buffer is supposed to be `TRACED_SYSCALLS_STR_LEN` long. Just re-use the `tracing_str` buffer above. This way, you'll have a single buffer for all the purposes.

include/tilck/kernel/test/tracing.h
```diff
@@ -0,0 +1,9 @@
+/* SPDX-License-Identifier: BSD-2-Clause */
+
+#pragma once
+
+#include <tilck/mods/tracing.h>
+STATIC bool simple_wildcard_match(const char *str, const char *expr);
+STATIC int set_traced_syscalls_int(const char *str);
+extern bool *traced_syscalls;
+STATIC extern char *traced_syscalls_str;
```
vvaltchev, line 9: "STATIC extern" does not make any sense. It's either one or the other. Please check the code.
Also, add a vertical space to separate the functions from the variables.

vvaltchev (changes requested):

vvaltchev:
Anyway, please rebase your changes on the top of the master branch. The build failures we're seeing here should be gone.

vvaltchev (changes requested):
Much better than before. Thanks for the fixes. There many more test cases that it would good to have, but that's a start. I've requested just two minor fixes and than PR will be good enough for merge. I'll add the rest of the tests myself and post a pull request.

**Important note:** the tag for tests is "[gtests]", not "[gtest]". Please change the commit message.

Issue #147, opened 2022-10-12
Add tests for save_param_buffer() and dump_param_buffer()


vvaltchev:
As it can be seen here: https://app.codecov.io/gh/vvaltchev/tilck/blob/master/modules/tracing/ptype_buffer.c, both the functions don't have enough coverage. 

It will be great to write unit tests for both the functions. Hints:
1.  Both the functions are `static` and need to be made `STATIC` in order to be used from unit tests
2.  Testing `save_param_buffer()` requires mocking `copy_str_from_user()` and `copy_from_user()`, in order to make them fail.
3. As an example of a similar issue, see: https://github.com/vvaltchev/tilck/issues/120

michelececcacci:
I'm taking this

rviu:
Hi, I would like to take this up

vvaltchev:
Hello @rviu, you're welcome to take this issue.

Issue #117, opened 2022-08-08
Add tests for register_on_task_exit_cb()


vvaltchev:
As you can see from codecov: https://app.codecov.io/gh/vvaltchev/tilck/blob/master/kernel/exit.c, the functions `register_on_task_exit_cb()` and `unregister_on_task_exit_cb()` don't have test coverage. It will be good to add a test for them. This can be done via a kernel self-test.

michelececcacci:
Mind if i take this?

vvaltchev:
> Mind if i take this?

Not at all, be my guest :-)

0xdeadbad:
Hey, just to update that I'm working on this one, life's been rough so it's been a slow process.

vvaltchev:
Sure, @0xdeadbad, no problem.

0xdeadbad:
@vvaltchev Hello, so I looked a lot at the kernel files, learnt a lot on how things are done and designed, but I'm still kinda stuck trying to write a test for this one. I really want to but I gotta kinda burnout with college and etc. I even noticed the code for `sb16.c` (audio device) that makes use of this functionality. I tried using kernel threads, but they differ from user threads I guess and don't call the callbacks set for its task. The callbacks are called in the `void terminate_process(..)` but there's an assertion the assure the current task is not a kernel thread, which makes things harder for me with my current knowledge. Could you give me a hint on this, if possible?

vvaltchev:
Hello @0xdeadbad, I'm glad that you did spend some time on that. Whole callbacks are designed to be used by the kernel, but are called when user tasks exist, so we both need an user process and a way to register such a "test" callback in the kernel.

First of all, the entry point of such test could be a function in `test_misc.c`. Don't forget to register that function (shellcmd) in the `cmds_table.h` file. From there, you could:

 1. fork() to create a child process
 2. read showhow (see below) a variable from the kernel
 3. tell somehow the kernel to register our test callback
 4. exit

Then, from the parent process, you should:
 1. wait for the child to exit
 2. read somehow (see below) the kernel variable and check that its value has increased

Now, one question is: **how to communicate with the kernel?** In general we have syscalls, but in this case we're talking about a test-specific interface. I've introduced a syscall for that called `sys_tilck_cmd()`. The idea is that each special test interface calls this syscall with a CMD parameter as 1st argument. At the top of the `tilck_cmd.c` file, you'll see a list of tilck cmds. 

[...]

PR #158, opened 2023-06-02
[gtests] Add tests for save_param_buffer() and dump_param_buffer()

rviu:
Closes https://github.com/vvaltchev/tilck/issues/147

### DIFF
--- a/tests/unit/tracerTest.cpp
+++ b/tests/unit/tracerTest.cpp
@@ -60,3 +60,52 @@ TEST(tracer_test, save_param_buffer)
    EXPECT_TRUE(save_param_buffer(data, data_sz, dest_buf_3, dest_bs_3));
    EXPECT_STREQ(dest_buf_3, "<fault>");
 }
+
+TEST(tracer_test, dump_param_buffer)
+{
+   MockingTracer mock;
+
+   ulong orig = 1;
+   long data_bs = -1;
+   long real_sz = -1;
+
+   char *data_1 = (char *)"\r";
+   char dest_1[10];
+   size_t dest_bs_1 = sizeof(dest_1);
+
+   char *data_2 = (char *)"\"";
+   char dest_2[10];
+   size_t dest_bs_2 = sizeof(dest_2);
+
+   char *data_3 = (char *)"\\";
+   char dest_3[10];
+   size_t dest_bs_3 = sizeof(dest_3);
+
+   EXPECT_TRUE(dump_param_buffer(orig, data_1, data_bs, real_sz, dest_1, dest_bs_1));
+   EXPECT_STREQ(dest_1, "\"\\r\"");
+
+   EXPECT_TRUE(dump_param_buffer(orig, data_2, data_bs, real_sz, dest_2, dest_bs_2));
+   EXPECT_STREQ(dest_2, "\"\\\"\"");
+
+   EXPECT_TRUE(dump_param_buffer(orig, data_3, data_bs, real_sz, dest_3, dest_bs_3));
+   EXPECT_STREQ(dest_3, "\"\\\\\"");
+
+   // For `if (dest_end - dest < ml - 1)` path
+   // and `if (dest >= dest_end - 4)` path
+   char *data_4 = (char *)"VeryVeryLong";
+   char dest_4[10];
+   size_t dest_bs_4 = sizeof(dest_4);
+
+   EXPECT_TRUE(dump_param_buffer(orig, data_4, data_bs, real_sz, dest_4, dest_bs_4));
+   EXPECT_STREQ(dest_4, "\"Very...\"");
+
+   // For `if (s == data_end && real_sz > 0 && data_bs < real_sz)` path
+   char *data_5 = (char *)"abcd";
+   char dest_5[10];
+   size_t dest_bs_5 = sizeof(dest_5);
+
+   real_sz = 2;
+
+   EXPECT_TRUE(dump_param_buffer(orig, data_5, data_bs, real_sz, dest_5, dest_bs_5));
+   EXPECT_STREQ(dest_5, "\"ab\"");
+}
--- a/tests/unit/tracerTest.cpp
+++ b/tests/unit/tracerTest.cpp
@@ -1,0 +1,62 @@
+/* SPDX-License-Identifier: BSD-2-Clause */
+
+#include <tilck/common/basic_defs.h>
+#include <gtest/gtest.h>
+#include <gmock/gmock.h>
+#include "mocking.h"
+
+using namespace testing;
+
+extern "C" {
+#include <tilck/common/string_util.h>
+#include <tilck/common/printk.h>
+
+#include <tilck/kernel/user.h>
+#include <tilck/mods/tracing.h>
+
+#include <tilck/kernel/test/tracing.h>
+}
+
+class MockingTracer : public KernelSingleton {
+public:
+
+   MOCK_METHOD(int, copy_str_from_user, (void *, const void *, size_t, size_t *), (override));
+   MOCK_METHOD(int, copy_from_user, (void *, const void *, size_t), (override));
+};
+
+TEST(tracer_test, save_param_buffer)
+{
+   MockingTracer mock;
+
+   long data_sz = -1;
+   void *data = (void *)"test";
+
+   char dest_buf_1[8];
+   size_t dest_bs_1 = sizeof(dest_buf_1);
+
+   char dest_buf_2[8];
+   size_t dest_bs_2 = sizeof(dest_buf_2);
+
+   char dest_buf_3[8];
+   size_t dest_bs_3 = sizeof(dest_buf_3);
+
+   EXPECT_CALL(mock, copy_str_from_user)
+      .WillOnce(Return(-1))
+      .WillOnce(Return(1));
+
+   // rc < 0
+   EXPECT_TRUE(save_param_buffer(data, data_sz, dest_buf_1, dest_bs_1));
+   EXPECT_STREQ(dest_buf_1, "<fault>");
+
+   // rc > 0
+   EXPECT_TRUE(save_param_buffer(data, data_sz, dest_buf_2, dest_bs_2));
+   EXPECT_EQ(dest_buf_2[dest_bs_2 - 1], 0);
+
+   data_sz = 0;
+
+   EXPECT_CALL(mock, copy_from_user)
+      .WillOnce(Return(1));
+
+   EXPECT_TRUE(save_param_buffer(data, data_sz, dest_buf_3, dest_bs_3));
+   EXPECT_STREQ(dest_buf_3, "<fault>");
+}
--- a/tests/unit/tracerTest.cpp
+++ b/tests/unit/tracerTest.cpp
@@ -32,27 +32,33 @@ TEST(tracer_test, save_param_buffer)
    void *data = (void *)"test";
 
    char dest_buf_1[8];
-   size_t dest_bs_1 = sizeof(dest_buf_1);
+   const size_t dest_bs_1 = sizeof(dest_buf_1);
 
    char dest_buf_2[8];
-   size_t dest_bs_2 = sizeof(dest_buf_2);
+   const size_t dest_bs_2 = sizeof(dest_buf_2);
 
    char dest_buf_3[8];
-   size_t dest_bs_3 = sizeof(dest_buf_3);
+   const size_t dest_bs_3 = sizeof(dest_buf_3);
 
    EXPECT_CALL(mock, copy_str_from_user)
       .WillOnce(Return(-1))
-      .WillOnce(Return(1));
+      .WillOnce([] (void *dest, const void *user_ptr, size_t, size_t *) {
+            strcpy((char *)dest, (char *)user_ptr);
+            return 1;
+         });
 
    // rc < 0
    EXPECT_TRUE(save_param_buffer(data, data_sz, dest_buf_1, dest_bs_1));
    EXPECT_STREQ(dest_buf_1, "<fault>");
 
    // rc > 0
-   EXPECT_TRUE(save_param_buffer(data, data_sz, dest_buf_2, dest_bs_2));
-   EXPECT_EQ(dest_buf_2[dest_bs_2 - 1], 0);
+   void *data_2 = (void *)"VeryVeryLong";
 
-   data_sz = 0;
+   EXPECT_TRUE(save_param_buffer(data_2, data_sz, dest_buf_2, dest_bs_2));
+   EXPECT_STREQ(dest_buf_2, "VeryVer");
+
+   // data_sz >= 0
+   data_sz = 5;
 
    EXPECT_CALL(mock, copy_from_user)
       .WillOnce(Return(1));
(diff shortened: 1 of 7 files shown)


rviu:
Hi @vvaltchev, the setup for the ptype_buffer test has been done (which I hope is correct) but I'm still figuring out writing the actual tests. I'm thinking something along the lines of:
```cpp
TEST(ptype_buffer, save_param_buffer)
{
   MockingPtypeBuffer mock;

   void * data = nullptr;
   long data_sz = -1;
   char dest_buf[] = "";
   size_t dest_bs = 0;

   EXPECT_CALL(mock, copy_str_from_user)
      .Times(1);

   EXPECT_TRUE(save_param_buffer(data, data_sz, dest_buf, dest_bs));
}
```
Am I on the right track?

vvaltchev:
Hello @rviu,
I took a look at the skeleton of the change. That's definitively a good start.

Just, the idea is not calling `save_param_buffer()` with a NULL pointer and negative input size, but calling it with real parameters but with a mocked `copy_str_from_user()` implementation that fails. This way the following code path will be tested:

```C
if (rc < 0) {
   /* Reading from `data` caused a PAGE fault */
   memcpy(dest_buf, "<fault>", 8);
} else if (rc > 0) {
```

Clearly, the test should test both the positive case (`copy_str_from_user()` succeeds) and the negative one. In the negative case, it should check that the dest buffer begins with "<fault>".

**Note**: being this test related to a small part of the tracer, the unit test should be called something like `tracerTest.cpp` as `ptype_buffer_test.cpp` does not make any sense in the larger context of Tilck's kernel. The mock class also should be renamed accordingly.

Anyway, you started very well. Maybe check https://github.com/vvaltchev/tilck/issues/120 and the final version of the pull request: https://github.com/vvaltchev/tilck/pull/143/commits/40b74f9cfb776986457cffbf774b58c8ef423a74 which is similar and/or check the gmock manual (http://google.github.io/googletest/gmock_cook_book.html) to learn how to use gMock. I agree that's not very intuitive.

Vlad

rviu:
Thank you for the detailed reply!

I went through the resources that you linked and made some new changes to the test (only for `save_param_buffer`, `dump_param_buffer` is still left).
- I've set `data_sz = -1` contrary to what you said; I couldn't find another way for the `if (data_sz == -1)` path.
- For the negative case, I'm testing with
```cpp
EXPECT_STREQ(dest_buf_1, "<fault>");
```
instead of
> In the negative case, it should check that the dest buffer begins with "".

because that seemed correct to me.
- And for the positive case:
```cpp
EXPECT_EQ(dest_buf_2[dest_bs_2 - 1], 0);
```
- Changed the name of the test

Please let me know if it requires more changes. Thanks

tests/unit/tracerTest.cpp
```diff
@@ -0,0 +1,63 @@
...
+using namespace testing;
+
+extern "C" {
+#include <tilck/common/string_util.h>
+#include <tilck/common/printk.h>
+
+#include <tilck/kernel/user.h>
+#include <tilck/mods/tracing.h>
+
+bool save_param_buffer(void *, long, char *, size_t);
```
vvaltchev, line 17: Add those declarations in `include/tilck/kernel/test/tracing.h` and include that header here. That header is (or should be) also included in the actual tracer's code. The purpose of the headers in kernel/test is precisely to avoid a signature mismatch error.

tests/unit/tracerTest.cpp
```diff
@@ -0,0 +1,63 @@
...
+   MOCK_METHOD(int, copy_str_from_user, (void *, const void *, size_t, size_t *), (override));
+   MOCK_METHOD(int, copy_from_user, (void *, const void *, size_t), (override));
+};
+
+TEST(tracer_test, save_param_buffer)
+{
+   MockingTracer mock;
+
+   long data_sz = -1;
+   void * data = (void *)"test";
```
vvaltchev, line 33: Style: no space between `*` and `data`.  Note: the cast for "test" is not really required in C.

vvaltchev (changes requested):
Very good, thanks for the work @rviu !! A few things to fix and I'll merge the pull request.

If you wanna keep both the commits, you should rename the .cpp test file there and all the other things internally so that the diff in the new commit simply implements the tests. Otherwise, squash the commits. 

Also, the commits needs to be renamed to follow the style of the rest of the commits and give more context to the reader. For example, the first could be something like: `[gtests] Intr. skeleton for tracer unit tests`. And the second could be: `[gtests] Intr. test for save_param_buffer()`.

Final note: the issue mentioned also the `dump_param_buffer()` function. You could implement a test for it in the 2nd commit or add a third one, as you prefer.

Thanks,
Vlad

vvaltchev:
> Thank you for the detailed reply!
> 

You're welcome!
Btw, I totally missed this reply this morning when I reviewed your new changes.

> I went through the resources that you linked and made some new changes to the test (only for `save_param_buffer`, `dump_param_buffer` is still left).
> 
> * I've set `data_sz = -1` contrary to what you said; I couldn't find another way for the `if (data_sz == -1)` path.

Yeah, I was talking in general without focusing on that specific parameter.
Note that there is also the code path where `data_sz` >= 0 that needs to be tested as well. The unit test should verify that `copy_from_user()` is called once and, if it failed, the dest should begin with `"<fault>"`.

> * For the negative case, I'm testing with
> 
> ```c++
> EXPECT_STREQ(dest_buf_1, "<fault>");
> ```
> 
> instead of
> 
> > In the negative case, it should check that the dest buffer begins with "".
> 
> because that seemed correct to me.

That's correct. I just noticed that writing anything that looks like an html tag in double-quotes  (e.g. `"<fault>"`) turns out into an empty quote, unless backquotes are used. That's weird. I did write `"<fault>"` but Github for some (maybe some security-related reason) decided to filter it out.

> 
> * And for the positive case:
> 
> ```c++
> EXPECT_EQ(dest_buf_2[dest_bs_2 - 1], 0);
> ```
> 
> * Changed the name of the test
> 
> Please let me know if it requires more changes. Thanks

I pointed a few more things to fix in today's review. But we're converging very fast :-)

Thanks,
Vlad

tests/unit/tracerTest.cpp
rviu: Without the cast, the following error is thrown:
```shell
error: invalid conversion from ‘const void*’ to ‘void*’ [-fpermissive]
   32 |    void *data = "test";
      |                 ^~~~~~
      |                 |
      |                 const void*

```

rviu:
Made a few more changes:
- Added the declarations in `include/tilck/kernel/test/tracing.h`
- Corrected style for `void *data`
- Added test for `dump_param_buffer()`
- Squashed the previous two commits

> Note that there is also the code path where `data_sz` >= 0 that needs to be tested as well. The unit test should verify that `copy_from_user()` is called once and, if it failed, the dest should begin with `"<fault>"`.

Is the existing test not sufficient for testing that path?
```cpp
   data_sz = 0;

   EXPECT_CALL(mock, copy_from_user)
      .WillOnce(Return(1));

   EXPECT_TRUE(save_param_buffer(data, data_sz, dest_buf_3, dest_bs_3));
   EXPECT_STREQ(dest_buf_3, "<fault>");
}
```
Please let me know if any more changes are required.

Thanks

tests/unit/tracerTest.cpp
```diff
@@ -0,0 +1,62 @@
...
+
+TEST(tracer_test, save_param_buffer)
+{
+   MockingTracer mock;
+
+   long data_sz = -1;
+   void *data = (void *)"test";
+
+   char dest_buf_1[8];
+   size_t dest_bs_1 = sizeof(dest_buf_1);
```
vvaltchev, line 35: `const size_t ...` is better as you're not changing those values later.

tests/unit/tracerTest.cpp
```diff
@@ -0,0 +1,62 @@
...
+      .WillOnce(Return(-1))
+      .WillOnce(Return(1));
+
+   // rc < 0
+   EXPECT_TRUE(save_param_buffer(data, data_sz, dest_buf_1, dest_bs_1));
+   EXPECT_STREQ(dest_buf_1, "<fault>");
+
+   // rc > 0
+   EXPECT_TRUE(save_param_buffer(data, data_sz, dest_buf_2, dest_bs_2));
+   EXPECT_EQ(dest_buf_2[dest_bs_2 - 1], 0);
```
vvaltchev, line 53: It's not really enough to test this case this way. See the code in `save_param_buffer()`:
```C
static bool
save_param_buffer(void *data, long data_sz, char *dest_buf, size_t dest_bs)
{
   if (data_sz == -1) {

      /* assume that `data` is a C string */
      int rc = copy_str_from_user(dest_buf, data, dest_bs, NULL);

      if (rc < 0) {

         /* Reading from `data` caused a PAGE fault */
         memcpy(dest_buf, "<fault>", 8);

      } else if (rc > 0) {

         /*
          * The user buffer is bigger than our reserved space: just truncate it.
          */
         dest_buf[dest_bs - 1] = 0;
      }
```
if `copy_str_from_user()` returned > 0, we expect the dest buffer to be truncated, but it needs to also have some content. The current test just checks that the dest buffer has been truncated. To make it really work, in one of the cases copy_str_from_user() should have an actual implementation as strcpy(), instead of just returning 1.

To do that, see "Using Functions/Methods/Functors/Lambdas as Actions" in http://google.github.io/googletest/gmock_cook_book.html. Briefly, you can pass a lambda to WillOnce, or use Invoke() and then pass a class + method. 

In alternative, if that's end up taking too much time, at least initialize the dest buffer with something well known and then check that the content it's the same but truncated.

tests/unit/tracerTest.cpp
```diff
@@ -0,0 +1,62 @@
...
+
+   data_sz = 0;
```
vvaltchev, line 55: If data is "test", here data_sz should be 5, not zero. Just think about testing the code in general, instead of using "tricks" to make it go in the flow you wanted. In other words, write tests that cover the abstract functionality of the code, without knowing all the details of its implementation.

vvaltchev (changes requested):
Thanks for the update, @rviu. I spent a bit more time thinking about this test so I added more suggestions in the last review. The same comments would apply for the 2nd commit, even if I didn't have enough time to deeply review that test. Just, keep in mind the general idea: write tests that cover all the code paths by thinking of what the function should do _in abstract_ for a given input, instead of thinking about the concrete implementation. In other words, don't think about triggering the code paths in order to increase the coverage _directly_, but make it happen as _naturally_ as possible by testing several inputs + different behavior for the mocks. This way, not only the tests will be generic and robust against implementation changes, but will detect much more defects. 100% line coverage != everything is tested.

Trivial example: how would a good test look like for an sqrt() function? It will be easy to write a test that covers 100% of the lines of an sqrt() implementation, but that allows sqrt(25) to be 3, for example.

I hope you like my inputs on testing. Let me know :-)

rviu:
Thanks for the review and the suggestions @vvaltchev.

> write tests that cover all the code paths by thinking of what the function should do _in abstract_ for a given input, instead of thinking about the concrete implementation

> Trivial example: how would a good test look like for an sqrt() function? It will be easy to write a test that covers 100% of the lines of an sqrt() implementation, but that allows sqrt(25) to be 3, for example.

Really good example! Helped me to better understand testing and how the current test was falling short.

I've made some more changes based on your feedback:
- Added a lambda function for `strcpy` in mock `copy_str_from_user`
- Test the contents of `dest_buf` accordingly, using a longer test string.

> The same comments would apply for the 2nd commit

I spent some time reviewing that test as well and it seems to me that it is accurately testing the corresponding function.

> I hope you like my inputs on testing. Let me know :-)

Absolutely, your inputs have been really helpful to me; nudging me in the right direction towards better quality code!

As always, please let me know if any more changes are required.

Thanks

vvaltchev:
Thank you for the fixes, @rviu. 
I'm so glad to hear that you appreciated my input :-)

vvaltchev:
@rviu, let me know if you'd like to make more contributions to the project: there are many interesting things that can be done here, both high-level and low-level, in C, C++, Python and Bash.

rviu:
@vvaltchev I'd love to contribute more to the project. Although I'm familiar mostly with C++, I wouldn't mind exploring other technologies :)

vvaltchev:
@rviu I'm so happy to hear that. Do you want to continue working on the unit tests? There is a fair amount of refactoring needed, plus many more new tests could be written. For example, it would be great to have full coverage in the unit tests for the Tilck console. I could prepare the skeleton for such tests, if you're interested in working on that.

Also, if you'd like to explore a bit Tilck in general and learn how to write other kind of tests, you might also take a look on this issue: https://github.com/vvaltchev/tilck/issues/117. Nobody is working on it at the moment.

If you like, maybe spend some time digging here and there, playing with the project and the various types of testing (https://github.com/vvaltchev/tilck/blob/master/docs/testing.md) so that you can get an idea of what is more interesting to you.

Of course, there are plenty of non-test related things to do both in the kernel and outside. Two quick examples:

- supporting aarch64 as HOST architecture for building Tilck (scripting / build engineering)
- writing a more powerful alternative to `ccmake(1) — cmake-curses-gui` in Python with ncurses
- porting Tilck to x86-64: the most challenging & rewarding task

rviu:
Porting Tilck to x86-64 seems pretty cool. I read the discussion page at https://github.com/vvaltchev/tilck/discussions/84 which mentions about it briefly. Could you give me a little more information about this task?

vvaltchev:
@rviu Well, that is a monumental "hardcore" kernel task. It requires making everything (kernel and userspace apps) to build and run on x86-64. I already started the work to make it at least simply build for the x86-64 target. That works now, but it's very far away from booting. 

**Detail about booting:** First of all, if we keep the current boot logic using multiboot, it will mean that the bootloader(s) will load Tilck in Protected Mode 32 as now, but the kernel will enter in Long Mode by itself. That could be OK for the moment, and re-evaluate it in the future. Probably it might make sense to switch to a different boot protocol (e.g. Linux boot protocol), because QEMU doesn't support directly boot loading with `-kernel` a 64-bit image. So, temporarily we'll need to make the x86-64 tests always do a a full boot using Tilck's bootloader, but that's limiting because it will be harder from the outside to pass boot options.

**Major changes required:**
All the arch/i386 files need to be ported to x86-64. In some cases, code can be shared and moved to x86-family, but in other cases the code needs to be completely rewritten.

- The paging needs to be re-written for x86-64 to support 4-levels or at least 3-levels + 1 hardcoded page directory

- The interrupt controller driver has to be re-written and use the APIC. Clearly, the old code needs to work as well and abstractions needs to be implemented such that generic code can use either of the interrupt controllers.

- The context switching has to be re-written to work for x86-64

- The whole GDT configuration has to change as x86-64 has a limited support for segmentation. 

- The support for TLS has to be re-written to work on x86-64

- The double-fault handling has to be made to work on x86-64

- Many syscalls and structures change a little bit from i386 to x86-64, therefore a whole new syscall table has to be created. Many wrappers will be required to make the same implementation work both for 32-bit and 64-bit Tilck

- A new syscall interface with the "SYSCALL" instruction needs to be supported

- All the userspace apps need to be compiled with a 64-bit toolchain and work correctly on 64-bit Tilck

- Many other smaller fixes that I cannot predict at the moment

This will probably require 3-6 months of full time work, even you have previous experience with kernel development. If you don't, it will take more time, I don't know how long. The Intel Software Developer's manual has to become your best friend: https://www.intel.com/content/dam/www/public/us/en/documents/manuals/64-ia-32-architectures-software-developer-vol-3a-part-1-manual.pdf

If you're able to do all of that, you'd be qualified to apply to professional positions as a kernel engineer in big companies as it will be considered a remarkable achievement. But the effort invested will need to be as equally exceptional. I could guide you a bit here and there, but 99% of the debugging and hard-work will need to be done on your side. Even if I'd like to spend some that working on that myself, at the moment I don't have enough time due to my full-time job and family commitments. So, at the moment I could act as a maintainer/reviewer/mentor for this project.

If you say you're really passionate about becoming a kernel developer and investing years in that field, that's absolutely the right task for you and I'll be happy to help at my best. Just, depending on your background and skills, we might need to start with something simpler first, as there's is much to learn. Getting something like that done is very different from typical userspace C++ development. 

Let me know more about your interests so we can figure out a plan. Also, I guess it would make sense to move this conversation out of Github. I don't have your email as it's not visible in your commits. So, if you want, drop me an email on the address you see in my commits, so we could continue talking about that.

Looking forward to talk with you,
Vlad

PR #160, opened 2023-07-21
Implement support for UEFI Runtime Services

rviu:


boot/efi/multiboot.c
```diff
@@ -7,6 +7,7 @@
 
 static multiboot_memory_map_t *multiboot_mmap;
 static UINT32 mmap_elems_count;
+static tilck_extra_boot_info *extra_boot_info;
```
vvaltchev, line 10: Why a pointer? You'd need to make it point to some valid mem area. That would require an allocation etc.
Just make it a plan object, as you'll going to make the actual allocation later anyways, and copy this object there.

boot/efi/multiboot.c
```diff
@@ -295,10 +296,38 @@ MbiSetPointerToAcpiTable(void)
     *       bootloader under QEMU: that's slower for tests and limiting for
     *       debugging purposes.
     */
-   gMbi->apm_table = (u32)tablePaddr;
+   // gMbi->apm_table = (u32)tablePaddr;
+   
+   extra_boot_info->RSDP = (u32)tablePaddr;
```
vvaltchev, line 301: That's good, but the whole comment above it doesn't apply here anymore. It should be moved to the place where `gMbi->apm_table` is set, because there is were we're going the "hack".

boot/efi/multiboot.c
```diff
@@ -295,10 +296,38 @@ MbiSetPointerToAcpiTable(void)
...
+   EFI_STATUS status = EFI_SUCCESS;
+   EFI_PHYSICAL_ADDRESS paddr = EFI_MBI_MAX_ADDR;
+
+   status = BS->AllocatePages(AllocateMaxAddress,
+                              EfiLoaderData,
+                              1,
+                              &paddr);
+   HANDLE_EFI_ERROR("AllocatePages");
+
+   BS->CopyMem(TO_PTR(paddr), extra_boot_info, sizeof(extra_boot_info));
```
vvaltchev, line 324: After making `extra_boot_info` a regular obj, not a pointer, fix this:
```
BS->CopyMem(TO_PTR(paddr), &extra_boot_info, sizeof(extra_boot_info));
```

boot/efi/multiboot.c
```diff
@@ -295,10 +296,38 @@ MbiSetPointerToAcpiTable(void)
...
+   gMbi->apm_table = (u32)paddr;
```
vvaltchev, line 325: Yep, here it goes the big fat comment about the "hack" we're doing.

include/system_headers/multiboot.h
```diff
@@ -269,6 +269,13 @@ struct multiboot_apm_info
   multiboot_uint16_t dseg_len;
 };
 
+typedef struct
+{
+  multiboot_uint32_t RSDP;
+  // EFI_RUNTIME_SERVICES *RT;
+  void *RT;
```
vvaltchev, line 276: This struct cannot go here, because the multiboot header is not Tilck's. It's 3rd-party code ("system_headers"). We do not touch such headers, unless it's impossible to make the project compile.

This struct should go in a dedicated in the common area between the kernel, the bootloader and everything else. So, make it be: `include/tilck/common/boot.h`.

Also, if you notice, there are no "void *" pointers in the multiboot header for a precise reason: the size has to be the same no matter what (the bitness of the bootloader vs the kernel might differ). Also, the data might not be accessible directly as in userspace. When dealing with the barrier between the bootloader and the kernel, addresses are typically integers as they cannot be simply de-referenced, unless there's a mapping in kernel's virtual space or paging is entirely disabled.

Therefore, just use `uint32` for both the values. Note that when Tilck is ported to x86-64, this will remain and continue to work: if you check the call to AllocatePages(), we're passing a MAX_ADDRESS which is  very low (EFI_MBI_MAX_ADDR = 64 KB). 

Finally, as per Tilck's coding style, no typedef.

kernel/arch/i386/CMakeLists.txt
```diff
@@ -1,6 +1,34 @@
 # SPDX-License-Identifier: BSD-2-Clause
 cmake_minimum_required(VERSION 3.2)
 
+# GNUEFI
```
vvaltchev, line 4: As discussed via e-mail, there's no need to attempt linking gnu-efi to the kernel. We only need structs and defines. It's only necessary the kernel to see the efiapi.h include file, as I mentioned in the email.

kernel/main.c
```diff
@@ -200,10 +200,11 @@ read_multiboot_info(void)
       const char *name = TO_PTR(mbi->boot_loader_name);
 
       if (!strcmp(name, "TILCK_EFI")) {
+         tilck_extra_boot_info* extra_boot_info = (tilck_extra_boot_info*)mbi->apm_table;
```
vvaltchev, line 203: This will make the 64-bit build to fail, because `apm_table` is 32-bit. We cannot directly cast a 32-bit integer into a pointer. This needs a double cast. So, just use `TO_PTR(mbi->apm_table)`.

Also, coding style: in Tilck the "*" goes next to the variable, not next to the type as in modern C++. The reason? If you declare `int* a, b;` you'll have just `a` as pointer, not `b`.

vvaltchev (changes requested):

include/tilck/common/boot.h
```diff
@@ -0,0 +1,8 @@
+#define EFI_RUNTIME_CODE MULTIBOOT_MEMORY_RESERVED+1
```
vvaltchev, line 1: Very Good. The only thing is that the names of those two defines is a bit misleading as they appear to be some UEFI constants, while they are totally ours. Therefore, could you please prefix them with "TILCK_BOOT_" so that's clear?

boot/efi/multiboot.c
```diff
@@ -63,9 +66,13 @@ EfiToMultibootMemType(UINT32 type)
...
 
       case EfiReservedMemoryType:
+         return MULTIBOOT_MEMORY_RESERVED;
+
       case EfiRuntimeServicesCode:
+         return EFI_RUNTIME_CODE;
+
       case EfiRuntimeServicesData:
-         return MULTIBOOT_MEMORY_RESERVED;
+         return EFI_RUNTIME_DATA;
```
vvaltchev, line 75: Correct, modulo the names. See the other comment about that.

kernel/arch/i386/CMakeLists.txt
```diff
@@ -4,6 +4,8 @@ cmake_minimum_required(VERSION 3.2)
 # Remove -rdynamic
 set(CMAKE_SHARED_LIBRARY_LINK_C_FLAGS)
 
+include_directories(${TCROOT}/i386/gnu-efi/inc)
```
vvaltchev, line 7: Correct, that should be all we need in the kernel.

kernel/main.c
```diff
@@ -200,10 +205,44 @@ read_multiboot_info(void)
...
       if (!strcmp(name, "TILCK_EFI")) {
+         struct tilck_extra_boot_info *extra_boot_info = TO_PTR(mbi->apm_table);
 
          printk("Multiboot: detected the TILCK_EFI bootloader\n");
-         printk("Multiboot: ACPI root ptr: %p\n", TO_PTR(mbi->apm_table));
-         acpi_set_root_pointer(mbi->apm_table);
+         printk("Multiboot: ACPI root ptr: %p\n", TO_PTR(extra_boot_info->RSDP));
+         acpi_set_root_pointer(extra_boot_info->RSDP);
+
+         /*
```
vvaltchev, line 214: OK, those are clearly experiments. Not OK for a merge, but it's fine to keep them while you're actively working on that.

kernel/mm/system_mmap.c
```diff
@@ -538,7 +539,9 @@ linear_map_mem_region(struct mem_region *r, ulong *vbegin, ulong *vend)
    const ulong pend = MIN((ulong)(r->addr+r->len), (ulong)LINEAR_MAPPING_SIZE);
 
    const bool rw = (r->type == MULTIBOOT_MEMORY_AVAILABLE) ||
-                   (r->extra & MEM_REG_EXTRA_KERNEL);
+                   (r->extra & MEM_REG_EXTRA_KERNEL) ||
+                   (r->type == EFI_RUNTIME_CODE) ||
```
vvaltchev, line 543: We don't need to map the EFI_RUNTIME_CODE in read/write mode. Just the data. The code should be mapped read-only.

vvaltchev (changes requested):
Very good progress! Almost everything was correct.

One thing: could you please squash all of your changes when you update the pull request? It does not make sense to have incremental commits that are wrong per se and having that fixed in another commit. Please check my details comments about that in this other pull-request: https://github.com/vvaltchev/tilck/pull/161

Briefly, either all the commits are squashed and there's just 1 commit per pull request or you spend quite some time rewriting the history of all the commits like it's required for series of patches for the Linux kernel to create a "nice story". The 1st approach is totally OK for Tilck and it's easy to do. But, if you really want to prepare for the Linux kernel, we could that as well, but it's a lot of extra work. No need to do that right now, unless that's a primary goal of yours. Again, please check my conversation with 0xdeadbad to get a full-picture of the trade-offs: https://github.com/vvaltchev/tilck/pull/161

Thanks,
Vlad

boot/efi/multiboot.c
```diff
@@ -144,6 +151,11 @@ MultibootSaveMemoryMap(UINTN *mapkey)
    do {
 
       UINT32 type = EfiToMultibootMemType(desc->Type);
+
+      if (desc->Attribute & EFI_MEMORY_RUNTIME) {
```
vvaltchev, line 155: That's a dirty hack. If the `desc->Type` is not enough for `EfiToMultibootMemType()` to work properly, then pass to it also the `desc->Attribute` and determine the type correctly there.

include/tilck/common/boot.h
```diff
@@ -0,0 +1,27 @@
...
+ * provided with a virtual address for calling
+ * SetVirtualAddressMap(). See
+ * setup_uefi_runtime_services() for more
+ * details.
+ */
+
+#define TILCK_BOOT_EFI_RUNTIME_CODE MULTIBOOT_MEMORY_RESERVED+1
+#define TILCK_BOOT_EFI_RUNTIME_DATA MULTIBOOT_MEMORY_RESERVED+2
+
+#define TILCK_BOOT_EFI_RUNTIME_CODE_REQ MULTIBOOT_MEMORY_RESERVED+3
```
vvaltchev, line 20: Why do we need to have this CODE_REQ and DATA_REQ types? Couldn't we just map them to the EFI_RUNTIME_CODE and DATA types? I haven't check the spec myself. I mean, do we need to keep mapped ALL of those region types or just the REQ ones? If it's just the REQ ones, we should totally skip the first two. If we need them all, why it's so important to distinguish them? 

Btw, for the sake of the tests, until everything works, keep all of those mem types mapped. It will be easier to remove them later and check if that's still OK.

kernel/main.c
```diff
@@ -200,10 +208,15 @@ read_multiboot_info(void)
...
       if (!strcmp(name, "TILCK_EFI")) {
+         struct tilck_extra_boot_info *extra_boot_info = TO_PTR(mbi->apm_table);
 
          printk("Multiboot: detected the TILCK_EFI bootloader\n");
-         printk("Multiboot: ACPI root ptr: %p\n", TO_PTR(mbi->apm_table));
-         acpi_set_root_pointer(mbi->apm_table);
+         printk("Multiboot: ACPI root ptr: %p\n", TO_PTR(extra_boot_info->RSDP));
+         acpi_set_root_pointer(extra_boot_info->RSDP);
+
+         uefi_rt_addr = extra_boot_info->RT;
```
vvaltchev, line 217: Not sure if we're going to use UEFI here in this file or move all the code that calls UEFI to a separate file. So, until we're playing with experiments, that's OK. We'll think about that detail later.

kernel/main.c
```diff
@@ -359,6 +418,9 @@ kmain(u32 multiboot_magic, u32 mbi_addr)
    init_kmalloc();
    init_paging();
 
+   if (booting_with_uefi)
```
vvaltchev, line 421: Cosmetic, but too ugly: don't check for this boolean here. Call the function unconditionally and check the boolean there (and return if it's false).

kernel/main.c
```diff
@@ -340,6 +353,52 @@ async_init(void)
       panic("Unable to create a kthread for do_async_init()");
 }
 
+static void
+setup_uefi_runtime_services() {
```
vvaltchev, line 357: Tilck should be portable as much as possible. Probably, this function should be moved to a dedicated file and be renamed something a bit more generic as "setup_firmware_runtime_services". But, since nothing is working at the moment, that's fine. Let's keep it as a note for later.

kernel/main.c
```diff
@@ -340,6 +353,52 @@ async_init(void)
...
+   EFI_RUNTIME_SERVICES *RT = TO_PTR(uefi_rt_addr);
```
vvaltchev, line 358: As mentioned in the email, there's no way to access this pointer that way. You'll have to use KERNEL_PA_TO_VA().

kernel/main.c
```diff
@@ -340,6 +353,52 @@ async_init(void)
...
+
+   struct mem_region ma;
+
+   EFI_MEMORY_DESCRIPTOR *virt_map = kmalloc((size_t)get_mem_regions_count());
+   int num_entries_virt_map = 0;
+
+   for (int i = 0; i < get_mem_regions_count(); i++) {
+      get_mem_region(i, &ma);
+
+      if (ma.type == TILCK_BOOT_EFI_RUNTIME_CODE_REQ || ma.type == TILCK_BOOT_EFI_RUNTIME_DATA_REQ) {
```
vvaltchev, line 370: Yeah, if only those memory types are really needed, we might just drop the other ones.

vvaltchev: Also, cosmetic. Avoid too many indentations by keeping the main code path outside an `if`. Just:
```
if (ma.type != CODE && ma.type != DATA)
   continue;
```

kernel/main.c
vvaltchev, line 357: Coding style:
1. In C, functions that don't accept parameters are declared like `foo(void)`. If you skip the `void`, it means that any number of params can be passed. That's different from C++.
2. The '{' goes on the new line.

kernel/main.c
```diff
@@ -340,6 +353,52 @@ async_init(void)
...
 
+static void
+setup_uefi_runtime_services() {
+   EFI_RUNTIME_SERVICES *RT = TO_PTR(uefi_rt_addr);
+
+   printk("Runtime Services: %p\n", RT);
+
+   struct mem_region ma;
+
+   EFI_MEMORY_DESCRIPTOR *virt_map = kmalloc((size_t)get_mem_regions_count());
```
vvaltchev, line 364: It doesn't matter for the sake of the experiment, but for the final change don't waste memory: first count the number of regions with a dedicated loop, then allocate the map and then fill it.

kernel/main.c
```diff
@@ -340,6 +353,52 @@ async_init(void)
       panic("Unable to create a kthread for do_async_init()");
 }
 
+static void
+setup_uefi_runtime_services() {
+   EFI_RUNTIME_SERVICES *RT = TO_PTR(uefi_rt_addr);
+
+   printk("Runtime Services: %p\n", RT);
+
+   struct mem_region ma;
```
vvaltchev, line 362: Cosmetic: declare the stack variables in the main scope at the beginning on the function. That allows with a quick look to get an idea of how much stack space a function is used. Declaring variables sparsely in the body is more "polluted" and doesn't help in C because we don't have RAII as in C++. Writing C code might look like a more "low-level" version of C++, but in reality is a whole different world. Different mindset etc. 

Sadly, often C++ programmers don't understand C programmers and vice-versa and end up almost "hating" each other because they believe the others are doing "stupid" things. While in reality, the "others" are living in a completely different world.

kernel/main.c
```diff
@@ -340,6 +353,52 @@ async_init(void)
...
+      if (ma.type == TILCK_BOOT_EFI_RUNTIME_CODE_REQ || ma.type == TILCK_BOOT_EFI_RUNTIME_DATA_REQ) {
+         EFI_MEMORY_DESCRIPTOR desc;
+
+         desc.Type = ma.type == TILCK_BOOT_EFI_RUNTIME_CODE_REQ ? EfiRuntimeServicesCode : EfiRuntimeServicesData;
+         desc.PhysicalStart = (ulong)ma.addr;
+         desc.VirtualStart = (ulong)KERNEL_PA_TO_VA(ma.addr);
+         desc.NumberOfPages = ma.len / PAGE_SIZE;
+         desc.Attribute = EFI_MEMORY_RUNTIME;
+
+         virt_map[num_entries_virt_map] = desc;
```
vvaltchev, line 379: Cosmetic:
```
virt_map[num_entries++] = desc;
```

kernel/main.c
```diff
@@ -340,6 +353,52 @@ async_init(void)
...
+
+   printk("Time %p\n", RT->GetTime);
+   /*
+   printk("SVAM %p\n", RT->SetVirtualAddressMap);
+
+   UINTN mem_map_size = num_entries_virt_map * sizeof(EFI_MEMORY_DESCRIPTOR);
+   UINTN desc_size = sizeof(EFI_MEMORY_DESCRIPTOR);
+   UINT32 desc_version = EFI_MEMORY_DESCRIPTOR_VERSION;
+
+   EFI_STATUS status = RT->SetVirtualAddressMap(mem_map_size, desc_size, desc_version, virt_map);
```
vvaltchev, line 396: Unfortunately, after the last conversation, it looks like that calling `SetVirtualAddressMap()` like this here won't work because the UEFI runtime mem regions are not identity-mapped. 

You might need to call `map_pages()` in the loop, for each of those regions, with VADDR = PADDR. See how `linear_map_mem_region()` uses that function. Then, here at the end, after the call succeeded, you'll need to re-iterate that loop and call: unmap_pages() for each region.

kernel/main.c
```diff
@@ -340,6 +353,52 @@ async_init(void)
...
+static void
+setup_uefi_runtime_services() {
+   EFI_RUNTIME_SERVICES *RT = TO_PTR(uefi_rt_addr);
+
+   printk("Runtime Services: %p\n", RT);
+
+   struct mem_region ma;
+
+   EFI_MEMORY_DESCRIPTOR *virt_map = kmalloc((size_t)get_mem_regions_count());
+   int num_entries_virt_map = 0;
```
vvaltchev, line 365: Cosmetic: no need for such long names for short-lived local variables. This is not BIG_ENTERPRISE_CODE (tm) :-) 

Something like `num_entries` is more than enough.

vvaltchev (changes requested):

include/tilck/common/boot.h
rviu: You are correct, we can skip the first two types. I tried doing that in the latest diff and it works.

kernel/main.c
rviu: I'm slowly beginning to understand this difference in mindsets, it's fascinating

kernel/main.c
```diff
@@ -41,11 +42,18 @@
...
+#define EFI_DEBUG 0
+#include <efi.h>
+#include <efiapi.h>
+
 static bool read_multiboot_info_passed;
 static u32 saved_multiboot_magic;
 static multiboot_info_t *saved_multiboot_mbi;
 static void read_multiboot_info(void);
 
+static bool booting_with_uefi;
```
vvaltchev, line 54: After some thinking, I really don't like putting UEFI stuff here like that. It's much better to create a kernel/uefi.c file and put everything there. Also, the `booting_with_uefi` is not necessary. You can simply check for `if (!uefi_rt_addr)` in the code.

vvaltchev: Also, introduce the `include/tilck/kernel/uefi.h` header file for putting all the declarations for the uefi-related functions. For the coding style, check the other headers.

kernel/main.c
```diff
@@ -340,6 +353,116 @@ async_init(void)
       panic("Unable to create a kthread for do_async_init()");
 }
 
+static void
+setup_uefi_runtime_services(void)
```
vvaltchev, line 357: Move this code to the new kernel/uefi.c file.

kernel/main.c
```diff
@@ -340,6 +353,116 @@ async_init(void)
...
+{
+   struct mem_region ma;
+
+   if (!booting_with_uefi)
+      return;
+
+   EFI_RUNTIME_SERVICES *RT = TO_PTR(uefi_rt_addr);
```
vvaltchev, line 364: Move to this variable declaration on the top. You can assign it's value later. Again, I know that in C++ that would be an abomination.

kernel/main.c
```diff
@@ -340,6 +353,116 @@ async_init(void)
       panic("Unable to create a kthread for do_async_init()");
 }
 
+static void
+setup_uefi_runtime_services(void)
+{
+   struct mem_region ma;
+
+   if (!booting_with_uefi)
```
vvaltchev, line 361: You can replace this check with `if (!uefi_rt_addr)`.

kernel/main.c
```diff
@@ -340,6 +353,116 @@ async_init(void)
...
+      return;
+
+   EFI_RUNTIME_SERVICES *RT = TO_PTR(uefi_rt_addr);
+   printk("Runtime Services: %p\n", RT);
+
+   int num_entries = 0;
```
vvaltchev, line 367: Move it to the top, next to the others.

kernel/main.c
```diff
@@ -340,6 +353,116 @@ async_init(void)
...
+
+   /*
+    * When SetVirtualAddressMap() is called, the physical
+    * addresses should be directly accessible. So, identity
+    * map them.
+    */
+   for (int i = 0; i < get_mem_regions_count(); i++) {
+      get_mem_region(i, &ma);
```
vvaltchev, line 375: Please add an empty line before this call, coding style: https://github.com/vvaltchev/tilck/blob/master/docs/contributing.md

kernel/main.c
```diff
@@ -340,6 +353,116 @@ async_init(void)
...
+
+      if (ma.type != TILCK_BOOT_EFI_RUNTIME_CODE && ma.type != TILCK_BOOT_EFI_RUNTIME_DATA)
```
vvaltchev, line 377: Line too long, probably way over 80 cols. You can transform it this way:
```
if (ma.type != TILCK_BOOT_EFI_RUNTIME_CODE &&
    ma.type != TILCK_BOOT_EFI_RUNTIME_DATA)
{
   continue;
}
```

Note the braces and the fact that the opening brace is on a new-line, contrary to all the other cases. That's because the expression has been split: https://github.com/vvaltchev/tilck/blob/master/docs/contributing.md

kernel/main.c
```diff
@@ -340,6 +353,116 @@ async_init(void)
...
+
+      ulong *vbegin = TO_PTR(pbegin);
+
+      /*
+       * SetupVirtualAddressMap() requires the runtime data
+       * as well as runtime code to be writable. So use
+       * `PAGING_FL_RW` for both.
+       */
+      size_t count =
+         map_pages(get_kernel_pdir(),
```
vvaltchev, line 393: OK, that works, but it believe it will be much better to have the first loop simply count the number of entries and nothing else. Map the pages in the next loop, before or assigning the descriptor. This loop could be simple as:
```
for (int i = 0; i < get_mem_regions_count(); i++) {

   if (ma.type == TILCK_BOOT_EFI_RUNTIME_CODE ||
       ma.type == TILCK_BOOT_EFI_RUNTIME_DATA)
   {
      num_entries++;
   }
}
```

kernel/main.c
```diff
@@ -340,6 +353,116 @@ async_init(void)
...
+                   PAGING_FL_BIG_PAGES_ALLOWED |
+                   PAGING_FL_RW);
+
+      if (count != page_count)
+         panic("kmalloc: unable to map regions in the virtual space");
+
+      num_entries++;
+   }
+
+   EFI_MEMORY_DESCRIPTOR *virt_map = kmalloc((size_t)num_entries);
```
vvaltchev, line 406: declare the variable at the top of the function

kernel/main.c
```diff
@@ -340,6 +353,116 @@ async_init(void)
...
+   num_entries = 0;
+
+   for (int i = 0; i < get_mem_regions_count(); i++) {
+      get_mem_region(i, &ma);
```
vvaltchev, line 410: empty line

kernel/main.c
```diff
@@ -340,6 +353,116 @@ async_init(void)
...
+
+      if (ma.type != TILCK_BOOT_EFI_RUNTIME_CODE && ma.type != TILCK_BOOT_EFI_RUNTIME_DATA)
```
vvaltchev, line 412: line too long, break the `if` as:
```
if (ma.type != TILCK_BOOT_EFI_RUNTIME_CODE &&
    ma.type != TILCK_BOOT_EFI_RUNTIME_DATA)
{
   continue;
}
```

kernel/main.c
```diff
@@ -340,6 +353,116 @@ async_init(void)
...
+         continue;
+
+      EFI_MEMORY_DESCRIPTOR desc;
```
vvaltchev, line 415: declare that at the beginning of the scope (= beginning of the for loop)

kernel/main.c
```diff
@@ -340,6 +353,116 @@ async_init(void)
...
+
+      desc.Type = (ma.type == TILCK_BOOT_EFI_RUNTIME_CODE) ? EfiRuntimeServicesCode : EfiRuntimeServicesData;
+      desc.PhysicalStart = (ulong)ma.addr;
+      desc.VirtualStart = (ulong)KERNEL_PA_TO_VA(ma.addr);
+      desc.NumberOfPages = ma.len / PAGE_SIZE;
+      desc.Attribute = EFI_MEMORY_RUNTIME;
+
+      virt_map[num_entries++] = desc;
+
+      printk("Type: %d\tPhysical Address: %lld\t Virtual Address: %lld\n", desc.Type, desc.PhysicalStart, desc.VirtualStart);
```
vvaltchev, line 425: There is no place for these debug printk() calls in the final change.

kernel/main.c
```diff
@@ -340,6 +353,116 @@ async_init(void)
...
+   }
+
+   printk("Virtual map: %p\n", virt_map);
```
vvaltchev, line 428: As above, no debug printk() calls.

kernel/main.c
```diff
@@ -340,6 +353,116 @@ async_init(void)
...
+   UINTN mem_map_size = num_entries * sizeof(EFI_MEMORY_DESCRIPTOR);
+   UINTN desc_size = sizeof(EFI_MEMORY_DESCRIPTOR);
+   UINT32 desc_version = EFI_MEMORY_DESCRIPTOR_VERSION;
+
+   EFI_STATUS status = RT->SetVirtualAddressMap(mem_map_size, desc_size, desc_version, virt_map);
+
+   printk("%d\n", status);
+
+   EFI_TIME Time;
+   RT->GetTime(&Time, NULL);
```
vvaltchev, line 442: Very important check missing: does this work AFTER the identity mapping has been removed? That's where it matters to work.

kernel/main.c
```diff
@@ -359,6 +482,8 @@ kmain(u32 multiboot_magic, u32 mbi_addr)
    init_kmalloc();
    init_paging();
 
+   setup_uefi_runtime_services();
+
```
vvaltchev, line 486: No extra line between `setup_uefi_runtime_services();` and `acpi_mod_init_tables();`

kernel/main.c
```diff
@@ -340,6 +353,116 @@ async_init(void)
...
+       * SetupVirtualAddressMap() requires the runtime data
+       * as well as runtime code to be writable. So use
+       * `PAGING_FL_RW` for both.
+       */
+      size_t count =
+         map_pages(get_kernel_pdir(),
+                   vbegin,
+                   pbegin,
+                   page_count,
+                   PAGING_FL_BIG_PAGES_ALLOWED |
```
vvaltchev, line 397: `unmap_pages()` doesn't support the BIG PAGES, so don't set this flag here.
Also, why do you pass always the RW flag? The code pages should be RO mapped.

kernel/main.c
```diff
@@ -340,6 +353,116 @@ async_init(void)
...
+
+   EFI_TIME Time;
+   RT->GetTime(&Time, NULL);
+
+   printk("Year %d, Month %d, Day %d\n", Time.Year, Time.Month, Time.Day);
+
+   for (int i = 0; i < get_mem_regions_count(); i++) {
+      get_mem_region(i, &ma);
+
+      if (ma.type != TILCK_BOOT_EFI_RUNTIME_CODE && ma.type != TILCK_BOOT_EFI_RUNTIME_DATA)
```
vvaltchev, line 449: Same coding style as above. Empty space after the `for`, split the `if` etc.

kernel/main.c
```diff
@@ -340,6 +353,116 @@ async_init(void)
...
+   EFI_STATUS status = RT->SetVirtualAddressMap(mem_map_size, desc_size, desc_version, virt_map);
+
+   printk("%d\n", status);
+
+   EFI_TIME Time;
+   RT->GetTime(&Time, NULL);
+
+   printk("Year %d, Month %d, Day %d\n", Time.Year, Time.Month, Time.Day);
+
+   for (int i = 0; i < get_mem_regions_count(); i++) {
```
vvaltchev, line 446: Before this loop, you should:
1. poison the `virt_map` array with memset().
2. free the `virt_map` heap object, hoping it will work. Very likely it will, but please try that.

It makes no sense to keep this heap object around forever if UEFI doesn't need it.

kernel/main.c
```diff
@@ -340,6 +353,116 @@ async_init(void)
...
+      const size_t page_count = (pend - pbegin) >> PAGE_SHIFT;
+
+      ulong *vbegin = TO_PTR(pbegin);
+
+      unmap_pages(get_kernel_pdir(),
+                  vbegin,
+                  page_count,
+                  false);
+   }
+}
```
vvaltchev, line 464: Here, at the very end, you should try (but not part of the change, just for tests) to call GetTime() and see if it works properly.

vvaltchev (changes requested):
A bunch of cosmetic/coding-style related changes + two major things:
1. check that GetTime() works after the final unmap()
2. check that GetTime() still works after unmap() & (poison+free) of the `virt_map`

vvaltchev:
Oops, I forgot another thing. For the final change, the commit message should be something like "[uefi] Add support for Runtime Services in the kernel". Optionally, you could add a nice description of how the change works in the commit message. If you don't want to do it, that's still fine, no problem.

include/tilck/kernel/uefi.h
```diff
@@ -0,0 +1,7 @@
+/* SPDX-License-Identifier: BSD-2-Clause */
+
+#pragma once
+
+extern ulong uefi_rt_addr;
```
vvaltchev, line 5: As mentioned in the previous review, please let's avoid exporting global variables everywhere. Encapsulation matters even in a low-level C projects like this one. Let's just export a `uefi_set_rt_pointer()` function that just sets that variable once, after asserting that's NULL (e.g. `ASSERT(!uefi_rt_addr)`. Also, that variable can now be made static inside the uefi.c translation unit.

kernel/main.c
```diff
@@ -200,10 +202,13 @@ read_multiboot_info(void)
...
       if (!strcmp(name, "TILCK_EFI")) {
+         struct tilck_extra_boot_info *extra_boot_info = TO_PTR(mbi->apm_table);
 
          printk("Multiboot: detected the TILCK_EFI bootloader\n");
-         printk("Multiboot: ACPI root ptr: %p\n", TO_PTR(mbi->apm_table));
-         acpi_set_root_pointer(mbi->apm_table);
+         printk("Multiboot: ACPI root ptr: %p\n", TO_PTR(extra_boot_info->RSDP));
+         acpi_set_root_pointer(extra_boot_info->RSDP);
+
+         uefi_rt_addr = extra_boot_info->RT;
```
vvaltchev, line 211: Please call `uefi_set_rt_pointer()` here to save the pointer.

kernel/main.c
```diff
@@ -200,10 +202,13 @@ read_multiboot_info(void)
       const char *name = TO_PTR(mbi->boot_loader_name);
 
       if (!strcmp(name, "TILCK_EFI")) {
+         struct tilck_extra_boot_info *extra_boot_info = TO_PTR(mbi->apm_table);
 
          printk("Multiboot: detected the TILCK_EFI bootloader\n");
-         printk("Multiboot: ACPI root ptr: %p\n", TO_PTR(mbi->apm_table));
-         acpi_set_root_pointer(mbi->apm_table);
+         printk("Multiboot: ACPI root ptr: %p\n", TO_PTR(extra_boot_info->RSDP));
```
vvaltchev, line 208: Missing `printk()` about the RT pointer, as requested in the previous review. Add something like:
```
printk("Multiboot: UEFI RT ptr: %p\n", TO_PTR(extra_boot_info->RT));
```

kernel/uefi.c
```diff
@@ -0,0 +1,159 @@
...
+      }
+
+      desc.Type =
+         ma.type == TILCK_BOOT_EFI_RUNTIME_CODE
+            ? EfiRuntimeServicesCode
+            : EfiRuntimeServicesData;
+
+      desc.PhysicalStart = (ulong)ma.addr;
+      desc.VirtualStart = (ulong)KERNEL_PA_TO_VA(ma.addr);
+      desc.NumberOfPages = ma.len / PAGE_SIZE;
```
vvaltchev, line 105: `ma.len >> PAGE_SHIFT`

kernel/uefi.c
```diff
@@ -0,0 +1,159 @@
...
+      num_entries++;
+   }
+
+   virt_map = kmalloc((size_t)num_entries);
+   num_entries = 0;
+
+   for (int i = 0; i < get_mem_regions_count(); i++) {
+
+      EFI_MEMORY_DESCRIPTOR desc;
+
```
vvaltchev, line 89: unnecessary extra space between `desc` and `get_mem_region()`.

kernel/uefi.c
```diff
@@ -0,0 +1,159 @@
...
+      if (ma.type != TILCK_BOOT_EFI_RUNTIME_CODE &&
+          ma.type != TILCK_BOOT_EFI_RUNTIME_DATA)
+      {
+         continue;
+      }
+
+      const ulong pbegin = (ulong)ma.addr;
+      const ulong pend = MIN((ulong)(ma.addr+ma.len), (ulong)LINEAR_MAPPING_SIZE);
+
```
vvaltchev, line 145: just move the `page_count` next to these other two.

kernel/uefi.c
```diff
@@ -0,0 +1,159 @@
...
+      const size_t page_count = (pend - pbegin) >> PAGE_SHIFT;
```
vvaltchev, line 61: Move the `page_count` next to the other two variables and skip the extra space.

kernel/uefi.c
```diff
@@ -0,0 +1,159 @@
...
+#include <efiapi.h>
+
+ulong uefi_rt_addr;
+
+void setup_uefi_runtime_services(void)
+{
+   struct mem_region ma;
+   EFI_RUNTIME_SERVICES *RT;
+   EFI_MEMORY_DESCRIPTOR *virt_map;
+
```
vvaltchev, line 23: No need for this extra space here.

kernel/uefi.c
```diff
@@ -0,0 +1,159 @@
...
+       */
+      size_t count =
+         map_pages(get_kernel_pdir(),
+                   vbegin,
+                   pbegin,
+                   page_count,
+                   PAGING_FL_RW);
+
+      if (count != page_count)
+         panic("kmalloc: unable to map regions in the virtual space");
```
vvaltchev, line 78: This is not kmalloc. Just skip the "kmalloc: " prefix.

kernel/uefi.c
```diff
@@ -0,0 +1,159 @@
...
+            : EfiRuntimeServicesData;
+
+      desc.PhysicalStart = (ulong)ma.addr;
+      desc.VirtualStart = (ulong)KERNEL_PA_TO_VA(ma.addr);
+      desc.NumberOfPages = ma.len / PAGE_SIZE;
+      desc.Attribute = EFI_MEMORY_RUNTIME;
+
+      virt_map[num_entries++] = desc;
+
+      printk("Type: %d\tPhysical Address: %lld\t Virtual Address: %lld\n", desc.Type, desc.PhysicalStart, desc.VirtualStart);
```
vvaltchev, line 110: Here and below, when the change is ready, you'll need to remove all of those debug `printk()` calls.

kernel/uefi.c
```diff
@@ -0,0 +1,159 @@
...
+   }
+
+   printk("Virtual map: %p\n", virt_map);
+
+   printk("Time %p\n", RT->GetTime);
+   printk("SVAM %p\n", RT->SetVirtualAddressMap);
+
+   UINTN mem_map_size = num_entries * sizeof(EFI_MEMORY_DESCRIPTOR);
```
vvaltchev, line 118: Can't we skip all of those extra variables? Just declare `EFI_STATUS status` above and simply do:
```
status = RT->SetVirtualAddressMap(num_entries * sizeof(EFI_MEMORY_DESCRIPTOR),
                                  sizeof(EFI_MEMORY_DESCRIPTOR), 
                                  EFI_MEMORY_DESCRIPTOR_VERSION,
                                  virt_map);
```

After that, checking that `status` is a success is mandatory. We can panic in case of a failure.

vvaltchev (changes requested):
Good work, getting close to something mergeable, but more polishing required.
Note that some of the comments are the same as in the previous review.

vvaltchev:
@rviu I wanted to merge the PR so I clicked on "ready for review" but I realized that there are several small problems:
 - it fails to merge because of a trivial conflict on `kernel/arch/i386/CMakeLists.txt`. Please rebase the change on master and solve the trivial conflict. I tried solving this with GitHub and that created the merge commit, but in any case is not enough because of the other problems.
 
 - the clang builds fail because of this error: 
 ```
/__w/1/s/kernel/uefi.c:117:38: error: implicit conversion changes signedness: 'int' to 'unsigned int' [-Werror,-Wsign-conversion]
   status = RT->SetVirtualAddressMap(num_entries * sizeof(EFI_MEMORY_DESCRIPTOR),
 ```

This can be fixed by making `num_entries` and unsigned. I use clang for additional extra safety checks that gcc doesn't have or doesn't implement in a convenient way. The `-Wconversion` is a very restrictive set of warnings that works well on clang, but it's unusable on GCC. If you're really curious, check this: https://vladonsoftware.wordpress.com/2019/02/12/the-wconversion-drama/

 - The 64-bit Tilck build (gcc_8_4_x86_64) fails with:
 ```
/__w/1/s/kernel/uefi.c:13:10: fatal error: efi.h: No such file or directory
 #include <efi.h>
 ```

Simply because you included gnu-efi only in `kernel/arch/i386/CMakeLists.txt`. You should include the x86-64 version of gnu-efi in the `kernel/arch/x86_64/CMakeLists.txt` file like this:
```
include_directories(${TCROOT}/x86_64/gnu-efi/inc)
```

This will work if I added the 64-bit gnu-efi in the container.  Just fix the PR and make GitHub run the checks. That's automatic if the PR is not marked as a draft. If the compilation still fails, maybe simply include the 32-bit gnu-efi, because the headers are the same, just to make the build pass. I'll fix the problem later. For the moment, nothing 64-bit in the kernel works, except for the build. It's valuable to at least ensure that.

rviu:
@vvaltchev 

> it fails to merge because of a trivial conflict on kernel/arch/i386/CMakeLists.txt. Please rebase the change on master and solve the trivial conflict. I tried solving this with GitHub and that created the merge commit, but in any case is not enough because of the other problems.

I'm unable to see the conflict that you are talking about. Are you asking for your newest commit (Merge branch 'master' into uefi) to be squashed with the previous [uefi] one?

tests/unit/CMakeLists.txt
```diff
@@ -14,6 +14,8 @@ include_directories(${CMAKE_SOURCE_DIR}/include/system_headers)
 include_directories(${GTEST_SOURCE_DIR}/include)
 include_directories(${GMOCK_SOURCE_DIR}/include)
 
+include_directories(${TCROOT}/i386/gnu-efi/inc)
```
vvaltchev, line 17: This wasn't supposed to go here but in `kernel/arch/x86_64/CMakeLists.txt` which is the CMake file for the 64-bit Tilck kernel. I have a build target for that and has nothing to do with the unit tests. Even if the 64-bit kernel cannot boot because a ton of things are missing, it is built.

vvaltchev (changes requested):
See the comment.

vvaltchev:
> @vvaltchev
> 
> > it fails to merge because of a trivial conflict on kernel/arch/i386/CMakeLists.txt. Please rebase the change on master and solve the trivial conflict. I tried solving this with GitHub and that created the merge commit, but in any case is not enough because of the other problems.
> 
> I'm unable to see the conflict that you are talking about. Are you asking for your newest commit (Merge branch 'master' into uefi) to be squashed with the previous [uefi] one?

No. Github showed me a conflict. I checked out manually your source and indeed when I did `git rebase master` I had to solve that conflict. That was caused by my change "Allow Tilck to build on Linux-aarch64" in which I touched a ton of files. But it's fine as you successfully rebased, from what I see. Don't worry.

tests/unit/CMakeLists.txt
vvaltchev: Oops, I noticed that now with your latest update the target `gcc_8_4_x86_64` succeeds. Maybe I didn't look at the error carefully? I believed that `gcc_8_4_x86_64` was broken because uefi.c couldn't find the header file. What was the problem?

rviu: I believe it was because the unit tests weren't able to find the gnu-efi header file

PR #162, opened 2023-09-01
Read hardware clock using UEFI Runtime Services + Latency test

rviu:


tests/self/se_time.c
```diff
@@ -157,3 +157,26 @@ void selftest_delay(void)
 }
 
 REGISTER_SELF_TEST(delay, se_manual, &selftest_delay)
+
+void selftest_latency(void)
+{
+   struct datetime d;
+   const int iters = 10000;
```
vvaltchev, line 164: 10k iters are not enough to stabilize the result. I tried it myself. The latency varies between 3k and 4k cycles.
That's not OK. You have to increase the iterations until the results vary a very little. A perf test like this lasts typically a few seconds. This runs immediately and doesn't produce stable results.

tests/self/se_time.c
```diff
@@ -157,3 +157,26 @@ void selftest_delay(void)
 }
 
 REGISTER_SELF_TEST(delay, se_manual, &selftest_delay)
+
+void selftest_latency(void)
```
vvaltchev, line 161: "latency" is not a meaningful name. What latency is measuring? 
I'd suggest something like "selftest_clock_latency".

tests/self/se_time.c
```diff
@@ -157,3 +157,26 @@ void selftest_delay(void)
...
+   printk("---------------------------------------------\n\n");
+
+   start = RDTSC();
+
+   for (int i = 0; i < iters; i++) {
+      hw_read_clock(&d);
+   }
+
+   duration = (RDTSC() - start) / iters;
+
```
vvaltchev, line 178: unnecessary extra line

tests/self/se_time.c
```diff
@@ -157,3 +157,26 @@ void selftest_delay(void)
...
+   printk("Latency: %" PRIu64 "\n", duration);
+}
+
+REGISTER_SELF_TEST(latency, se_short, &selftest_latency)
```
vvaltchev, line 182: Perf tests are never "se_short". It should last several seconds. But, when such tests are run in the cloud the VM can be preempted for a long time if it burns too many CPU cycles and the delays grow almost exponentially. Therefore a test that consume several seconds of CPU time, not several seconds of real time, needs to be se_long, otherwise will timeout sometimes in Azure Pipelines.

tests/self/se_time.c
```diff
@@ -157,3 +157,26 @@ void selftest_delay(void)
...
+
+   start = RDTSC();
+
+   for (int i = 0; i < iters; i++) {
+      hw_read_clock(&d);
+   }
+
+   duration = (RDTSC() - start) / iters;
+
+   printk("Latency: %" PRIu64 "\n", duration);
```
vvaltchev, line 179: "Latency" is not enough. After the value you should add "RDTSC cycles".

vvaltchev (changes requested):

tests/self/se_time.c
rviu: With 200k iters and 20 second test time, the latency varies < 2k cycles. Is this acceptable?

vvaltchev: @rviu Sorry, what do you mean with "the latency varies < 2k cycles" ? I understand that it falls below 2k, but how much it varies? A good-enough measurement varies by 2%. If it varies too much with too many iterations, we have to understand why. It's possible that it varies because of other threads, in particular the thread that reads the time intensively I talked about. So, I suggest putting: `disable_preemption()` and `enable_preemption()` around the loop and see if you can get more stable results with fewer iterations. The criteria is 2% variance, with fewer iterations as possible, of course.

rviu: > The latency varies between 3k and 4k cycles.

Oh, I just understood what you meant by this. I thought you were referring to how much the difference b/w maximum and minimum observed latency was for you.

It's completely different for me. For me the latency varies between 310,000 and 312,000. This is what I meant when saying < 2k cycles.

I'm running the self-test inside Tilck by running `selftest latency`.

vvaltchev: @rviu Aha, got it. Yeah, the actual number varies a lot depending on the environment. E.g. if you use KVM vs. pure software virtualization, it's usually a big difference.

20 seconds are a bit too much. Can you disable the preemption and bring down the number of iterations so that results are stable both with KVM and without it? It should be possible to get stable results in a few seconds.

rviu: @vvaltchev Now with 50k iters, it takes about 6 seconds using kvm and almost instantly with non kvm well within the variance criteria

vvaltchev: That's good, thank you!

PR #165, opened 2023-10-08
[64-bit] Use separate vaddrs for reading phy. mem. and for the kernel

rviu:
BASE_VA is used for reading the physical memory and KERNEL_VADDR is used for the kernel.

In 64-bit, we want to map the kernel at -2GB to be able to use 32-bit offsets. Hence, a separate vaddr was introduced for achieving this.

To avoid 64-bit specific tricks, this change is
generic and for both 32-bit and 64-bit kernels.

### DIFF
--- a/include/tilck/kernel/paging.h
+++ b/include/tilck/kernel/paging.h
@@ -32,8 +32,17 @@
  * These MACROs can be used for the linear mapping region in the kernel space.
  */
 
-#define KERNEL_PA_TO_VA(pa) ((void *) ((ulong)(pa) + KERNEL_BASE_VA))
-#define KERNEL_VA_TO_PA(va) ((ulong)(va) - KERNEL_BASE_VA)
+#if defined(TESTING) || defined(KERNEL_TEST)
+
+   #define KERNEL_PA_TO_VA(pa) ((void *) ((ulong)(pa) + BASE_VA))
+   #define KERNEL_VA_TO_PA(va) ((ulong)(va) - BASE_VA)
+
+#else
+
+   #define KERNEL_PA_TO_VA(pa) ((void *) ((ulong)(pa) + KERNEL_VADDR - KERNEL_PADDR))
+   #define KERNEL_VA_TO_PA(va) ((ulong)(va) - KERNEL_VADDR + KERNEL_PADDR)
+
+#endif
 
 extern char page_size_buf[PAGE_SIZE] ALIGNED_AT(PAGE_SIZE);
 extern char zero_page[PAGE_SIZE] ALIGNED_AT(PAGE_SIZE);
--- a/kernel/arch/i386/linker_script.ld
+++ b/kernel/arch/i386/linker_script.ld
@@ -8,8 +8,9 @@ SEARCH_DIR("=/tmp/x86-i686--glibc--stable/usr/i686-buildroot-linux-gnu/lib32");
 SEARCH_DIR("=/tmp/x86-i686--glibc--stable/usr/i686-buildroot-linux-gnu/lib");
 
 kernel_paddr        = @KERNEL_PADDR@;
-kernel_text_paddr   = kernel_paddr + 0x1000;
-kernel_va           = @KERNEL_BASE_VA@ + kernel_text_paddr;
+kernel_text_offset  = 0x1000;
+kernel_text_paddr   = kernel_paddr + kernel_text_offset;
+kernel_va           = @KERNEL_VADDR@ + kernel_text_offset;
 
 PHDRS
 {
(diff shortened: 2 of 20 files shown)


include/tilck/kernel/paging.h
```diff
@@ -32,8 +32,17 @@
  * These MACROs can be used for the linear mapping region in the kernel space.
  */
 
-#define KERNEL_PA_TO_VA(pa) ((void *) ((ulong)(pa) + KERNEL_BASE_VA))
-#define KERNEL_VA_TO_PA(va) ((ulong)(va) - KERNEL_BASE_VA)
+#if defined(TESTING) || defined(KERNEL_TEST)
```
vvaltchev, line 35: This is wrong. There is no need for this #ifdef and only `BASE_VA` needs to be used. The name of the macro is wrong now. It was good in the past when we used `KERNEL_BASE_VA`. Now, it should be something like `PA_TO_LIN_VADDR`. The other name too: `KERNEL_VA_TO_PA` should be `LIN_VADDR_TO_PA` or something like that.

kernel/arch/i386/linker_script.ld
```diff
@@ -8,8 +8,9 @@ SEARCH_DIR("=/tmp/x86-i686--glibc--stable/usr/i686-buildroot-linux-gnu/lib32");
 SEARCH_DIR("=/tmp/x86-i686--glibc--stable/usr/i686-buildroot-linux-gnu/lib");
 
 kernel_paddr        = @KERNEL_PADDR@;
-kernel_text_paddr   = kernel_paddr + 0x1000;
-kernel_va           = @KERNEL_BASE_VA@ + kernel_text_paddr;
+kernel_text_offset  = 0x1000;
+kernel_text_paddr   = kernel_paddr + kernel_text_offset;
```
vvaltchev, line 12: the `kernel_text_paddr` is unused now. No need to keep it.

kernel/arch/i386/paging_int.h
```diff
@@ -43,7 +43,7 @@
 
 #define PAGE_FAULT_FL_COW (PAGE_FAULT_FL_PRESENT | PAGE_FAULT_FL_RW)
 #define BIG_PAGE_SHIFT                                            22
-#define KERNEL_BASE_PD_IDX        (KERNEL_BASE_VA >> BIG_PAGE_SHIFT)
+#define KERNEL_BASE_PD_IDX               (BASE_VA >> BIG_PAGE_SHIFT)
```
vvaltchev, line 46: Incorrect macro name. The `KERNEL_` prefix is misleading now. This should be renamed to something like `BASE_VADDR_PD_IDX`.

kernel/arch/i386/start.S
```diff
@@ -23,7 +23,7 @@
                               MULTIBOOT_MEMORY_INFO |  \
                               MULTIBOOT_VIDEO_MODE)
 
-#define PAGE_DIR_PADDR ((offset page_size_buf) - KERNEL_BASE_VA)
+#define PAGE_DIR_PADDR ((offset page_size_buf) - KERNEL_VADDR + KERNEL_PADDR)
```
vvaltchev, line 26: No... you're using KERNEL_VADDR in too many place. ALWAYS anywhere KERNEL_VADDR should be replaced with `BASE_VA` simply because we don't care in all of these places the address of kernel's text! We care about the linear mapping of the physical memory in the virtual address space.

kernel/arch/i386/start.S
```diff
@@ -58,7 +58,7 @@ multiboot_entry:
 
    /*
     * Before jump to kernel, we have to setup a basic paging in order to map
-    * the first 4-MB both at 0 and at +KERNEL_BASE_VA. Using 4-MB pages.
+    * the first 8-MB both at 0 and at +(KERNEL_VADDR - KERNEL_PADDR), using 4-MB pages.
```
vvaltchev, line 61: wrong

kernel/arch/i386/start.S
```diff
@@ -83,9 +83,9 @@ multiboot_entry:
    mov dword ptr [edx + 0], MAKE_BIG_PAGE(0)
    mov dword ptr [edx + 4], MAKE_BIG_PAGE(4 * MB)
 
-   # Map the first 8 MB also at KERNEL_BASE_VA
-   mov dword ptr [edx + (KERNEL_BASE_VA >> 20) + 0], MAKE_BIG_PAGE(0)
-   mov dword ptr [edx + (KERNEL_BASE_VA >> 20) + 4], MAKE_BIG_PAGE(4 * MB)
+   # Map the first 8 MB also at +(KERNEL_VADDR - KERNEL_PADDR)
+   mov dword ptr [edx + ((KERNEL_VADDR - KERNEL_PADDR) >> 20) + 0], MAKE_BIG_PAGE(0)
```
vvaltchev, line 87: Wrong. It won't work when KERNEL_VADDR is very different than BASE_VA like the 64-bit case. See the comments above.

vvaltchev (changes requested):
Sorry man, but the change is not OK. In too many cases you still use KERNEL_BASE_VA instead of the new BASE_VA. Also, the change description is not accurate and doesn't describe the problem as it should. I'll send you a private email with more comments.

PR #168, opened 2023-10-23
[64-bit] Allow Tilck to be mapped at a non-linear vaddr

vvaltchev:
This is the part of two of commit 29ee48c7f3. Here we completely decouple the default linear mapping of the physical address space from the kernel's vaddr of text, data & bss. By default, the 32-bit Tilck kernel will continue to be mapped at BASE_VA + KERNEL_PADDR, using the general-purpose linear mapping, but when the KRN32_LIN_VADDR option is enabled, the kernel itself will be mapped at a much higher address, linear the LINEAR_MAPPING_END.

The reason is the same as 29ee48c7f3: the 64-bit kernel will need to be mapped well outside of the 4 GB limit in order to support more physical memory in the linear mapping. But, doing that while keeping the linear mapping will be extremely inefficient because all the major compilers will refuse to compile an ELF binary placed at such a high vaddr without using -mcmodel=large. And the problem with that is that -mcmodel=large uses inefficient instructions to access global variables and call functions because it doesn't assume that data and text are within 2 GB of any point. It assumes that each variable or function can be *anywhere* in the 64-bit virtual address space, while that is not true. In theory, it's totally possible to patch the compiler and make it use a small memory model at a high address, but why bothering with that? The mainstream solution is simply to place the kernel somewhere within [-2 GB, 0). This allows both using the smaller (and more efficient) memory model and to leave all the virtual space we want for user programs.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -263,6 +263,9 @@ set(KERNEL_64BIT_OFFT ON CACHE BOOL
 set(KRN_CLOCK_DRIFT_COMP ON CACHE BOOL
     "Compensate periodically for the clock drift in the system time")
 
+set(KRN32_LIN_VADDR ON CACHE BOOL
+    "Place the 32-bit kernel in the default linear mapping")
+
 # Kernel options (disabled by default)
 
 set(KRN_PAGE_FAULT_PRINTK OFF CACHE BOOL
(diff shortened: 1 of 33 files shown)


vvaltchev:
@rviu Take a look at this change.

PR #169, opened 2023-11-05
[kernel]: Use a switch statement in kopt_handle_bool()

rilysh:
Hello,
`kopt_handle_bool(...)` function uses `strcmp(...)` to compare a single character, but as it's just a single character and not a stream of characters, we shall not need to use the `strcmp(...)` function.

### DIFF
--- a/kernel/cmdline.c
+++ b/kernel/cmdline.c
@@ -43,14 +43,19 @@ struct kopt {
 
 static void kopt_handle_bool(bool *var, const char *arg)
 {
-   if (*arg == '-')
-      *var = true;
-   else if (!strcmp(arg, "0"))
+   switch (*arg) {
+   case '0':
       *var = false;
-   else if (!strcmp(arg, "1"))
+      break;
+
+   case '-':
+   case '1':
       *var = true;
-   else
+      break;
+
+   default:
       NOT_REACHED();
+   }
 }
 
 static void kopt_handle_long(long *var, const char *arg)


vvaltchev:
Hello @rilysh,
I appreciate the contribution but in this specific case I'm rejecting it. First of all, all of that code will run only once at boot, parsing the boot options. The minimal overhead of using `strcmp()` is totally negligeable in that case. Executing a few extra instructions, one time doesn't matter. It matters if that was in a code path that can be executed multiple times, after boot.

Said that, I would still accept the change if the new code was simpler to read while keeping the pre-existing behavior. In this case, note that the the code does NOT keep the pre-existing behavior because it checks ONLY for the *first* char of the argument string. `strcmp()` instead, checks that the string *ends there*. With your code "0abc" will be treated as "0" and that's not OK. If you start adding checks for the trailing '\0' you'll see that the code will become way more verbose than that. I did all of that myself, when I ended up using `strcmp()` at the end.

So yeah, Tilck is all about extra performance and micro-optimizations too.. but in this case I don't see anything any benefit from doing that change.

Regards,
Vlad

rilysh:
> With your code "0abc" will be treated as "0" and that's not OK. 

Indeed, I just couldn't able to find a case where more than a character is used for kopt_handle_bool that's why I just kept a single character check but I really underestimated it.

> If you start adding checks for the trailing '\0' you'll see that the code will become way more verbose than that. 

Yeah for each character, it'd be quite messy. 

Thanks for the explanation there! Sorry for the noise.

vvaltchev:
No problem, @rilysh. It's OK to try :-)

PR #171, opened 2023-12-06
[64-bit] Introduce a fake implementation for fault_resumable_call()

vvaltchev:
This allows the x86-64 Tilck to:

   - initialize the textmode_video term
   - initialize term
   - initialize the console
   - run panic() and display a message correctly


vvaltchev:
@rviu Take a look at this simple change. Now Tilck 64 can print something on the screen and crashes gracefully with a simple NOT_IMPLEMENTED panic. No #ifdefs.

PR #174, opened 2023-12-13
Implemented minimal version of getrusage with its relative devshell test

cozis:
Hi Vlad!

This is a minimal implementation of `getrusage` that only reports user and kernel time (`rusage.ru_ctime` and `rusage.ru_stime`) and sets the remaining fields to 0. I was also thinking of ways to report values for some of the remaining fields, but wanted to get your opinion first!


cozis:
Seems like the devshell test itself doesn't work for azure (it works locally for me). Maybe the kernel busy loop was a bad idea?

include/tilck/kernel/sys_types.h
```diff
@@ -69,6 +69,10 @@ struct k_utimbuf {
    long modtime;
 };
 
+#define RUSAGE_SELF     0
```
vvaltchev, line 72: At least for the moment, sys_types.h is supposed to just include the system headers and get all the defines from there. The structs and defines in this file are only for the special cases when we cannot just include the system files or when that causes problems because of kernel/userspace incompatibilities. In the future, we might change everything and remove all the includes of system headers. In that case, we'll place everything here. But for the moment, that's the convention.

include/tilck/kernel/syscalls.h
```diff
@@ -134,7 +134,8 @@ CREATE_STUB_SYSCALL_IMPL(sys_sigpending)
 CREATE_STUB_SYSCALL_IMPL(sys_sethostname)
 CREATE_STUB_SYSCALL_IMPL(sys_setrlimit)
 CREATE_STUB_SYSCALL_IMPL(sys_old_getrlimit)
-CREATE_STUB_SYSCALL_IMPL(sys_getrusage)
+
+int sys_getrusage(int who, struct k_rusage *user_buf);
```
vvaltchev, line 138: STYLE: can you move this close to the function below?

kernel/syscalls.c
```diff
@@ -152,6 +153,73 @@ ulong sys_times(struct tms *user_buf)
...
+{
+   struct task *curr = get_curr_task();
+   struct k_rusage buf;
+   clock_t utime_ticks;
+   clock_t stime_ticks;
+   struct k_timespec64 utime;
+   struct k_timespec64 stime;
+
+   if (who == RUSAGE_CHILDREN)
+      // TODO: Resource usage of a process's children
```
vvaltchev, line 175: STYLE: use /* */ comments like this:
```
/*
 * something something
 * something.
 */
```

kernel/syscalls.c
```diff
@@ -152,6 +153,73 @@ ulong sys_times(struct tms *user_buf)
...
+      //       isn't supported yet!
+      return -EINVAL;
+
+   if (who != RUSAGE_SELF && who != RUSAGE_THREAD)
+      return -EINVAL;
+
+   // Since there can only be one thread per process,
+   // RUSAGE_SELF and RUSAGE_THREAD have the same meaning.
+
+   disable_preemption();
```
vvaltchev, line 185: Not sure if disabling the preemption is enough. I guess that the ticks are touched in the IRQ 0 handler (timer). So, you need to disable the interrupts here.

kernel/syscalls.c
```diff
@@ -152,6 +153,73 @@ ulong sys_times(struct tms *user_buf)
...
+      .ru_oublock = 0,
+      .ru_msgsnd = 0,
+      .ru_msgrcv = 0,
+      .ru_nsignals = 0,
+      .ru_nvcsw  = 0,
+      .ru_nivcsw = 0,
+   };
+
+   if (copy_to_user(user_buf, &buf, sizeof(buf)) != 0)
+      return (int) -EFAULT;
```
vvaltchev, line 218: STYLE: no need for the `(int)` cast.

kernel/syscalls.c
```diff
@@ -152,6 +153,73 @@ ulong sys_times(struct tms *user_buf)
...
+      .ru_inblock = 0,
+      .ru_oublock = 0,
+      .ru_msgsnd = 0,
+      .ru_msgrcv = 0,
+      .ru_nsignals = 0,
+      .ru_nvcsw  = 0,
+      .ru_nivcsw = 0,
+   };
+
+   if (copy_to_user(user_buf, &buf, sizeof(buf)) != 0)
```
vvaltchev, line 217: STYLE: no need for the long form `! = 0`. The implicit form is prettier.

tests/system/test_misc.c
```diff
@@ -389,3 +389,88 @@ int cmd_exit_cb(int argc, char **argv)
...
+}
+
+#define BEGIN_OF_TIME (struct timeval) {0, 0}
+
+static void busy_wait(void)
+{
+   // It's assumed that this loop takes
+   // more than 1us of cpu time.
+   for (int i = 0; i < 1000000; i++)
+      asm(""); // Don't let the compiler optimize this loop
```
vvaltchev, line 408: `asm("")` can be optimized by the compiler. Only `asmVolatile` is not optimized. I'd recommend `asmVolatile("nop")` even if, in the future, we'll need to replace that with something more portable.
Also, 1,000,000 of iterations are a lot. We don't need to burn that many cycles for testing something so simple and linear as getrusage(). Can you reduce the number of iterations?

tests/system/test_misc.c
```diff
@@ -389,3 +389,88 @@ int cmd_exit_cb(int argc, char **argv)
...
+}
+
+static void busy_wait_kernel(void)
+{
+   // It's assumed that this loop takes
+   // more than 1us of cpu time in kernel
+   // space.
+   for (int i = 0; i < 100; i++) {
+      int n = fork();
```
vvaltchev, line 417: I appreciate the attempt, but this is the incorrect way of doing that.
1. Note that that the timer frequency (50-250 HZ), not the 1us precision of the timespec struct.
2. You're not waiting for all the child to exit before exiting from here. That very likely causes all the tests to fail: https://vkvaltchev.visualstudio.com/Tilck/_build/results?buildId=1959&view=logs&jobId=28f1ef74-9f3f-5899-b019-7358c934de06&j=b80d80ca-e3a4-5429-e1b2-036d7cb8925c&t=b6270386-105a-5b35-2e48-584241c3a271
3. fork() is not the right tool. Use a custom Tilck syscall wait kernel time. See TILCK_CMD_SYSCALL, sys_tilck_cmd and the kernel/tilck_cmd.c file. There, you can add an empty slot for a TILCK test syscall, and then set the right function pointer in modules/systests/systests.c with the implementation. 4. In the test TILCK syscall, just burn some cycles with a loop and NOP. But don't do it too much, we don't need to make tests unnecessarily slower.

tests/system/test_misc.c
```diff
@@ -389,3 +389,88 @@ int cmd_exit_cb(int argc, char **argv)
...
+   struct rusage buf_1;
+
+   // Do some work to make sure that the user cpu time
+   // isn't less than 1us
+   busy_wait();
+   busy_wait_kernel();
+
+   // Check that RUSAGE_SELF doesn't fail and that
+   // the returned utime and stime aren't 0.
+   rc = syscall(SYS_getrusage, RUSAGE_SELF, &buf_0);
```
vvaltchev, line 436: Don't use `syscall()` unless it's really necessary. Use the libc wrapper instead.

vvaltchev (changes requested):
Hi Francesco, thanks for working on this.
Even if it requires a few changes, great start! I appreciate the passion for the tests. I wouldn't even ask for so good tests for such a simple feature, but I appreciate them ;-)

I have no time for a full review at the moment, but I did my best in a short time and wanted to give you a quick feedback. Note that the CI/CD tests run for your change and are all red. See them in the "checks" tab and click in the links to see the tests runs on Azure Pipelines. Direct link: https://vkvaltchev.visualstudio.com/Tilck/_build/results?buildId=1959&view=results

To avoid that in the future, you need to first run all the tests on your machine with `build/st/run_all_tests -c`. See docs/testing.md.

include/tilck/kernel/sys_types.h
cozis: Okay. I didn't notice that it included the system headers. I'll remove the defines and include `sys/resource.h`. Now I understand the meaning of the `k_` prefix

include/tilck/kernel/syscalls.h
cozis: done!

kernel/syscalls.c
cozis: While writing this function I tried to mirror the implementation of `sys_times`, which only guards the access of the tick counts with `disable_preemption`/`enable_preemption`. Should `sys_times` change too or is there a reason I'm missing why they're not the same?

tests/system/test_misc.c
cozis: > Note that that the timer frequency (50-250 HZ), not the 1us precision of the timespec struct.

I don't think I understand. Are you saying I should consider the resolution of the timer instead of the `struct timeval`?

cozis:
Thank you SO MUCH for the feedback!! I greatly appreciate it and will fix these problems as soon as possible

kernel/syscalls.c
vvaltchev: @cozis Great observation! I quickly too a look at `sys_times()` and `sched_account_ticks()` which is called from the IRQ handler. sched_account_ticks() might not need to disable the interrupts because the IRQ timer won't nest in itself (there's a check), but for `sys_times()` an all the other places where the ticks are used, just preemption is disabled and that doesn't seem enough. Even if the a single 32-bit address can be accessed kind-of atomically on x86, accessing both `curr->ticks.total` and `curr->ticks.total_kernel` certainly it is not atomic. The timer IRQ handler can run between the two writes. If we want consistent results, better just disable the interrupts, here and in all the other places where those variables are accessed. sched_account_ticks() might be simply moved in the disabled interrupts block in the caller and assert that interrupts are disabled internally.

tests/system/test_misc.c
vvaltchev: `struct timeval` is the structure of the interface, but that doesn't mean that the operating system guarantees a resolution is small as what can be represented in that struct. Of course, not even Linux does. It depends on the HW and how the time spent is calculated. On newer machines (with "newer" I typically mean machines after 2004/2005, maybe 2008 not sure), it is possible to estimate more precisely the CPU time we spent running a task on a CPU thanks to the HPET. But Tilck still uses the most traditional and primitive way of calculating the time spent in tasks, which is based on the scheduler "ticks". Each tick happens with the same frequency, determined by the primary timer. Again, Linux now supports variable ticks, tickless mode etc. But, traditionally, that's the simplest thing it can be done. Therefore, with a freq of 250 HZ the smallest time quantum we can appreciate is 4 ms. Consider that's relatively good. On old machines in the early 90s, the freq was 100 HZ and that time was 10ms. 

Therefore, when writing loops like that, consider that you wanna burn at few quanta. Not less because otherwise there could be less than 1 tick, but not more either, not not waste resources. Also, in the test itself, consider that in Azure sometimes VMs get stuck or can run very slowly. Therefore, it's possible that a long loop can take much more time than on a regular machine, in some corner cases.

cozis:
Hello! I finally managed to fix all of the issues. It took more effort than I imagined!

There are a couple things I'd like to point out:
* When reading the tick counts from the task object I disabled interrupts instead of disabling preemption. I figured disabling interrupts implied disabling preemption.
* The symbols `RUSAGE_SELF` and `RUSAGE_CHILDREN` are now imported from `sys/resource.h`, but since `RUSAGE_THREAD` is an extension I also needed to define it in `kernel/sys_types.h` in case it wasn't already. Defining `_GNU_SOURCE` before the inclusions didn't seem to work on some builds.
* The devshell test now relies on the timeout to detect the test failure

kernel/tilck_cmd.c
```diff
@@ -63,6 +65,13 @@ static int sys_tilck_run_selftest(const char *u_selftest)
    return se_run(se);
 }
 
+static int sys_busy_wait(ulong n)
```
vvaltchev, line 68: Naming: this is not a Linux syscall. Therefore, its name shouldn't start with `sys_`. Make it be something like `tilck_busy_wait()`.

Btw, this whole function is set-only and it should be placed in modules/systests/systests.c and registered with `register_tilck_cmd()`. See the other "test" syscalls in systests.c.

vvaltchev (changes requested):
Great job, Francesco! The pull request is almost perfect: a few minor touches here and there, and it's ready.

One more thing: can you please squash all the commits into one? For the reasoning & context, check the comments in https://github.com/vvaltchev/tilck/pull/161. [Btw, I should copy-paste all of that in a nice document or extend contributing.md].

TL;DR: Briefly, unless the commit history for the pull request is perfectly curated with constant rebasing (Linux-kernel style), which is very time-consuming to do, so the 2nd most clean thing is to squash all the temporary changes and merge one clean commit. But, please read the whole discussion in pull request 161, because I believe it has some value.

kernel/syscalls.c
```diff
@@ -152,6 +153,75 @@ ulong sys_times(struct tms *user_buf)
    return (ulong) get_ticks();
 }
 
+static struct k_timeval 
+timespec_to_timeval(struct k_timespec64 ts)
```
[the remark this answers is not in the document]
cozis: I didn't find an equivalent function. Should I move it to tilck/kernel/datetime.h?

kernel/syscalls.c
```diff
@@ -152,6 +153,75 @@ ulong sys_times(struct tms *user_buf)
...
+      return -EINVAL;
+
+   if (who != RUSAGE_SELF && who != RUSAGE_THREAD)
+      return -EINVAL;
+
+   /* Since there can only be one thread per process,
+    * RUSAGE_SELF and RUSAGE_THREAD have the same meaning.
+    */
+   ulong var;
+   disable_interrupts(&var);
```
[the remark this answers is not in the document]
cozis: That feels much cleaner! I'll do that.

kernel/tilck_cmd.c
cozis: No problem! Though it's not clear to me the benefit in only loading these "tilck command" functions lazily. Is it to avoid loading their code?

cozis:
When everything looks good to you, I'm going to try cleaning up the history by rebasing. It feels like a good skill to have!

kernel/syscalls.c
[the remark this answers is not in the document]
vvaltchev: Yeah, this should go to `tilck/kernel/datetime.h` but we should be very careful about its name as there are several timespec and timeval structs. I'd suggest calling it `k_ts64_to_k_timeval()`  or `k_timespec_to_k_timeval()`, implying that a timespec is 64-bit by default. Something like that. The `k_` prefix matters to avoid namespace-clashing with the userspace structs, which are also included in the headers. Including those headers in the kernel is both good and bad. Good because we're implementing Linux compatibility at binary level anyway, so why re-defining everything? Also, including those headers forces us to use non-standard names when there is a clashing and that is good to avoid causing confusion and make searching of identifiers easier. But including the system headers is also bad because sometimes the userspace structs are different from the kernel-space ones and because some userspace structs have extra hidden (non-officially documented)  `__fields` depending on the libc (glibc vs musl for example). So, for all of those cases, we need to have a dedicated `k_` struct in the kernel. 

Just to share the whole story with you, I even developed a fancy plugin for the clang static analyzer that allows me to check if any struct in Tilck changed its binary layout when compiling the kernel using a different system root for the system headers. I haven't been using it for 2-3 years, but it was very useful to check that the Tilck doesn't expose to userspace structs that are different, depending on how it is compiled. We might need it again in the future.

tests/system/test_misc.c
```diff
@@ -389,3 +389,102 @@ int cmd_exit_cb(int argc, char **argv)
...
+{
+   // Arbitrary amount of work
+   for (int i = 0; i < 10000; i++)
+      asmVolatile("nop"); // Don't let the compiler optimize this loop
+}
+
+static void busy_wait_kernel(void)
+{
+   int rc = syscall(TILCK_CMD_SYSCALL, TILCK_CMD_BUSY_WAIT, 10000);
+   if (rc)
```
vvaltchev, line 411: I'd recommend just using `DEVSHELL_CMD_ASSERT(rc == 0)` here with a simple comment on the top like: `/* TILCK_CMD_BUSY_WAIT must never fail */`

tests/system/test_misc.c
```diff
@@ -389,3 +389,102 @@ int cmd_exit_cb(int argc, char **argv)
...
+
+      for (n = 0; !comes_after(buf_1.ru_stime, buf_0.ru_stime); n++) {
+
+         busy_wait_kernel();
+
+         rc = getrusage(RUSAGE_SELF, &buf_1);
+         DEVSHELL_CMD_ASSERT(rc == 0);
+         DEVSHELL_CMD_ASSERT(++n < max);
+      }
+
```
vvaltchev, line 464: Style: extra empty line

vvaltchev (changes requested):
Great job! We're really at the end. All of the comments are about cosmetic details. One more thing, commit's headline it's a bit too long. For example "with its relative devshell test" is very verbose. Maybe a shorter headline could be something like `[syscalls] Introduce sys_getrusage() + devshell tests` ? Check the a few commits with `git log` to have an idea. I'm using shorter headlines for commit messages to mimic a bit the Linux kernel ones. Plus, tags are used when possible for a better navigation in the future. Sometimes I abbreviate "Introduce" with "Intr." to save characters and fit in 74 cols or less, but that is not necessary here.

kernel/tilck_cmd.c
vvaltchev: @cozis Good question. The benefit is to not introduce test-related code in the kernel which not only increases the code size, but can be a security risk. If we imagine what "production Tilck" could be, will it have those test functions? It's much better to keep the non-essential code outside of the *core* kernel. That's why we have modules and the kernel self tests in `tests/self` are also compiled conditionally when KERNEL_SELFTESTS is 1. That way is super simple to build a minimal Tilck. Consider that not only tests, but all the non-essential code, including the console, the ACPI support, the PS/2 keyboard driver and other stuff are source modules (we don't have loadable modules at runtime).

**Important** point about the test functions: while your function is totally fine if ends up in production, what about `TILCK_CMD_CALL_FUNC_0` which allows userspace to call an arbitrary kernel function **without** any checks? The idea is that in the *test* syscalls we can do _ANYTHING_. No security, no checks to avoid crashes, nothing. Testing can do anything and that's correct. We can also do live-patching and any kind of tricks. But in the actual product, those things shouldn't exist.

In the actual kernel, we can only have *real* Tilck syscalls. So far we have none of such Tilck-only syscalls. Anything in `tilck_cmds` is test-related stuff.  Actually, now I'm thinking that ideally all the function pointers in the `tilck_cmds` array should be NULL. That's not the case only for historical reasons.

The modules are relatively new in Tilck. I introduced them when decided to experiment with the "minimal Tilck" and see how small this kernel get and continue to work. So far, the minimal configuration can run in a i686 QEMU VM with just 3 MB of RAM and have 500-600 KB free. That's good, but I can could make it smaller, if I removed 600 KB for `busybox` and maybe 60-70 KB by removing the self tests. Just, it wouldn't be testable, but for true embedded it will great. Once the whole thing starts to run on ARM64, then a porting to ARM32 might be necessary, as the Cortex series is 32-bit only I think. Then, we can super-creative. Consider that for true embedded we might completely skip the bootloader, busybox and a ton of stuff. To some degree, things will be easier and quite simpler, depending on what we wanna support. But I wanted to have a "friendly" OS that can also run "regular" stuff not just some micro-embedded real-time OS, that's why we're here, trying to do both.

kernel/syscalls.c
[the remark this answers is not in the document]
cozis: Right! I'll go with `k_ts64_to_k_timeval`. Thank you for the explanation, the whole thing wasn't very clear to me.

kernel/tilck_cmd.c
cozis: Got it. I wasn't thinking about the bigger picture but only about my function which is innocuous. 

Do you think that having modules that load at runtime in Tilck makes sense? Would adding them increase complexity too much?

tests/system/test_misc.c
cozis: Ok! I wrote that comment because usually I make a distinction between a test that failed and one that aborted.

tests/system/test_misc.c
```diff
@@ -389,3 +389,98 @@ int cmd_exit_cb(int argc, char **argv)
    DEVSHELL_CMD_ASSERT(after_cb == before_cb + 1);
    return 0;
 }
+
+static bool
+comes_after(struct timeval t2,
+            struct timeval t1)
+{
+   // t1 must come before t2
+   return t2.tv_sec > t1.tv_sec || (t2.tv_sec == t1.tv_sec && t2.tv_usec > t1.tv_usec);
```
vvaltchev, line 398: Style: this line is 87 chars long, while the hard limit is 80 (see docs/contributing.md). Can you please split it?

Btw, you can run the following script from project's root to check for long lines easily: `./scripts/dev/count_lines_of_code -l`.  This way, you can find and fix any other "long" lines exceeding 80 cols.  At the moment, there are 0 lines exceeding 80 cols in the project, and I'm trying to keep that enforced unless there is a really good case for making an exception.

vvaltchev (changes requested):
The change is practically ready. Just one or more lines to split to avoid exceeding 80-cols and squashing all the commits into one, or into 2 or 3, as we discussed offline. Whatever you prefer.

cozis:
Hei Vlad! I finally squished the commits and shortened the commit message. Also I made sure that the `./scripts/dev/count_lines_of_code -l` didn't report any lines

cozis:
In the end I decided to go with the one commit since rebasing seemed a little tricky for the time being. I'll try to go with the "model 2" in the next pull request!

vvaltchev:
Great job, @cozis!! :-)

PR #187, opened 2024-07-04
[fork]Fixed a critical issue in the fork system call execution path

aenrbes:
Hi @vvaltchev 

I was recently researching tilck and discovered that there is a criticality problem with fork system calls，
The cause of the problem is as I describe in the commit content.

Way to reproduce the problem：
Add a “for‘’ loop at the following code to allow an interrupt to occur that satisfies the two conditions that caused this problem，

```
diff --git a/kernel/arch/i386/arch_syscalls.c b/kernel/arch/i386/arch_syscalls.c
index 21afb73..0cfc6a2 100644
--- a/kernel/arch/i386/arch_syscalls.c
+++ b/kernel/arch/i386/arch_syscalls.c
@@ -581,6 +581,10 @@ static void do_syscall(regs_t *r)
 
    process_signals(curr, sig_pre_syscall, r);
    enable_preemption();
+
+   for (ulong i = 0; i < 1000000;i++)
+      ;
+
    {
       trace_sys_enter(sn,r->ebx,r->ecx,r->edx,r->esi,r->edi,r->ebp);
       do_syscall_int(fptr, r);
```
It can be reproduced immediately when tilck is started


vvaltchev:
@aenrbes Thanks for finding this bug and proposing a fix!
At a first look, it makes sense to me. Indeed, I was playing some time ago in the `bt` branch with
`KRN_MINIMAL_TIME_SLICE=1` and `TIMER_HZ=500` to stress the kernel and found out several defects reproducible mostly in CI/CD runs due to the greater timing variability. I think one of the bugs occurred during `fork()` and was probably exactly this one. I didn't have the time to continue digging into this but I'm happy that you did.

The fix looks very good now. However, I don't wanna rush merging it immediately and risk introducing yet another subtle defect. One of these days, I'll spend some time thinking about the implications of this change and the possibility of corner cases. I'll either merge the PR or request some changes, if there's something extra to fix.

Thanks again,
Vlad

kernel/arch/i386/arch_syscalls.c
```diff
@@ -602,6 +602,7 @@ void handle_syscall(regs_t *r)
     */
    r->eflags |= EFLAGS_IF;
 
+   get_curr_task()->user_state_regs = r;
```
vvaltchev, line 605: Even if we end up actually needing a new regs pointer like `user_state_regs`, that would be set in `save_current_task_state()` conditionally when we're in syscall context, but I'm not convinced we need that. See my other comments below.

kernel/fork.c
```diff
@@ -107,11 +107,11 @@ int do_fork(bool vfork)
    ASSERT(curr->state_regs != NULL);
 
    child->state_regs--; // make room for a regs_t struct in child's stack
-   *child->state_regs = *curr->state_regs; // copy parent's regs_t
+   *child->state_regs = *curr->user_state_regs; // copy parent's regs_t
```
vvaltchev, line 110: Here the preemption is disabled. So, at most an IRQ can happen. There are 3 cases:
1. An IRQ happens at _before_ this statement: that's not a problem because the stack will change and eventually the IRQ handler will return and everything will be correct as before.
2. An IRQ happens _after_ this statement: that's not a problem because we already copied the registers.
3. An IRQ happens _during_ this statement: **yep, that's a problem**. But that doesn't mean we need a separate `user_state_regs` field etc. The bug I see is simply that this copy operation is not IRQ safe. Therefore, the simplest thing we can do to fix this is to perform the copy while keeping the interrupts disabled:

```C
disable_interrupts_forced();
{
   *child->state_regs = *curr->state_regs;
}
enable_interrupts_forced();
```
Does this make sense or am I missing something else?
My general thinking is to avoid introducing "new state" in any form unless that's unavoidable. A bigger state means almost always "more complexity" and makes the whole thing harder to reason about. Of course, if you show me that my suggestion doesn't work and somehow IRQs before that point somehow change the state in an undesirable way, can talk about that.

The assumption that an IRQ restores the state as it was before is of **critical importance**. If that has a bug, let's fix it as well.

kernel/fork.c
```diff
@@ -107,11 +107,11 @@ int do_fork(bool vfork)
...
    set_return_register(child->state_regs, 0);
 
    // Make the parent to get child's pid as return value.
-   set_return_register(curr->state_regs, (ulong) child->tid);
+   set_return_register(curr->user_state_regs, (ulong) child->tid);
```
vvaltchev, line 114: If we can avoid the new `user_state_regs`, that's not necessary.

vvaltchev (changes requested):

kernel/fork.c
aenrbes: Because of the above, simply turning off the interrupt when the `*child->state_regs = *curr->state_regs;`statement is executed won't solve the problem.  So the user's regs state must be saved somewhere else

vvaltchev: @aenrbes Thank you for clarifying that. It wasn't obvious and I didn't spent enough time on the problem. 
So, in my understanding now what happens is that after that initial `enable_preemption()` in `do_syscall()`, the current task gets preempted (the IRQ 0 triggering that is irrelevant, the code could also have yield etc.). When that happens, `state_regs` points to the current _kernel_ state which is later used as the regs for the child, while `running_in_kernel=false` for the child task and that leads to any sort of fancy corruption during the task switch. While with your fix it works because we save the USER regs separately and make the new task work with that. Is that correct?

Now, if that's the case, my follow-up question for you is: will `user_state_regs` used only by `fork()` or it should play a role elsewhere?

* If it's only used by fork(), we could make fork() be a special syscall and be called by `do_special_syscall()`. That will allow us to say that we don't want preemption enabled by default for it. Then, since it's sub-optimal to introduce such a large non-preemptable code, we could add another abstraction to the syscall infra with a new flag like SYSFL_RAW_REGS and make sys_fork() and sys_vfork() to accept a `regs_t *` parameter. Then, the parameter will be forwarded to do_fork() and saved before enabling the preemption. The gain with this approach is avoiding introducing new per-task state and is used only temporarily by one syscall. [ If you don't wanna do all of that, you could simply flag `fork` and `vfork` as SYSFL_NO_PREEMPT and check if that solves the problem as it should in my current understanding. I'll do the follow-up optimization, if you don't want/have time ].

* On the other side, if there is another use case for that or if creating new use-cases for the new field will simplify something, then it makes sense add the new state. I'm not so sure about that and haven't spent enough time thinking about that as I'm out of home most of the time these days. What do you think?

I'm wondering at this point if something can go wrong with the preemption of user space tasks. In theory, it's all good because the new kernel stack and state are added on the top of the user space one:
```

+---------------+
|  IRQ handler  |
+---------------+
|  kernel funcs |     <- state_regs (kernel)
+---------------+
|  syscall      |
+---------------+
|  user regs    |    <- state_regs (user space)
+---------------+
```

That's the model I used for a long time after the "early days" when I kept a copy of the state registers explicitly in the struct task. That was pointless since they have been already saved on the stack. The new model allowed a long series of simplifications and performance improvements but hey.. I missed the fork() case. Do you see some other limitation to that model besides the specific fork() case?

aenrbes: You're absolutely right. user_state_regs is used only by fork(). In fact, the first solution I came up with was to pass the user_regs pointer as a function argument, just because it changes the code a little bit more.

And I don't see any other problems with the model of preemption of user space tasks at the moment.

I have modified the code using the function parameter method, tested it, and committed it below

vvaltchev: Thanks for re-working your change, @aenrbes. It looks great now.
I'm going to merge it.

PR #188, opened 2024-07-16
[term] disable preemption before safe_ringbuf_write_elem()

aenrbes:
Hi @vvaltchev

I noticed that it is forgot to add the mutex protection before calling `safe_ringbuf_write_elem()` in `term_execute_or_enqueue_action_template()`. 

I discovered this problem when this `panic("Read from safe_ringbuf interrupted on-going write. Not supported")` occurred during a long period of testing runall.

### DIFF
--- a/include/tilck/kernel/term_aux.h
+++ b/include/tilck/kernel/term_aux.h
@@ -58,7 +58,10 @@ term_execute_or_enqueue_action_template(term *t,
 {
    bool was_empty;
 
+   disable_preemption();
    if (UNLIKELY(!safe_ringbuf_write_elem(&rb_data->rb, a, &was_empty))) {
+
+      enable_preemption_nosched();
 
       term_handle_full_ringbuf(t,
                                rb_data,
@@ -67,6 +70,9 @@ term_execute_or_enqueue_action_template(term *t,
                                exec);
 
       /* NOTE: do not return */
+   } else {
+
+      enable_preemption_nosched();
    }
 
    if (!was_empty) {


include/tilck/kernel/term_aux.h
```diff
@@ -58,15 +58,20 @@ term_execute_or_enqueue_action_template(term *t,
 {
    bool was_empty;
 
+   kmutex_lock(&rb_data->lock);
```
vvaltchev, line 61: Grabbing the sleeping lock is inefficient just to write to the ring buffer which is a very fast operation. The overhead of that kind of lock is too high compared to the code is trying to protect. The idea behind the "safe_ringbuf" was it is safe against IRQs as long as on the stack there is only one consumer/reader and all the rest of the IRQs and nested IRQs are producers. See the long comment at the beginning of safe_ringbuf.h. 

It now it turns out that because of user space tests there is a way to violate the expectations in this code, maybe because of kernel preemption, then I guess we should use the basic ringbuf.h implementation and protect it by disabling the interrupts.

Do you mind checking a bit deeper what went wrong?

vvaltchev:
Thanks @aenrbes for your extensive testing and fixes! It's much appreciated.

I have some concerns for this case not only because of the overhead of the mutex, but because if that's unsafe, then the mutex should be grabbed also by the reader etc. So, I guess that some assumptions in that code are probably incorrect. In other words, I didn't forget grabbing the mutex in that case: I assumed that it was safe skipping that because of the "safe_ringbuf" assumptions, while apparently it wasn't probably due to kernel preemption. I believe I suspected that for some time, that's why I added some extra checks in safe_ringbuf. That's we saw the nice panic message instead of a weird corruption.

So probably an actual "IRQ lock" plus a the simpler ringbuf could solve the problem in a better way. (Tilck does not have a proper IRQ lock because does not support SMP, so we just disable/enable the interrupts, but I guess you got the idea).

In alternative, you could disable/enable the preemption when the ringbuf is accessed and keep the safe_ringbuf. Smaller change to do. Not sure if it's the best, but could do the job now if you don't wanna make a bigger change. 

I'm starting to think that I over-engineered the "safe_ringbuf" and its use-cases while trying to avoid too hard to disable/enable the interrupts. That might not be the right trade-off. On that line of thought, another possible solution is to make the "safe_ringbuf" much safer and simpler by making it a thin-wrapper over the basic ringbuf and disable/enable the interrupts while using it. 

What do you think about that, overall?

aenrbes:
In the following call stack, preemption is always on. So presumably the root cause is kernel preeption occurred while executing `safe_ringbuf_write_elem `in this call stack. My opinion is that for now it should be sufficient to use the method of disable preemption.


```
term_execute_or_enqueue_action_template(exec_action_func exec, struct term_action * a, struct term_rb_data * rb_data, term * t)
serial_term_execute_or_enqueue_action(struct sterm * t, struct term_action * a)
sterm_write(term * _t, const char * buf, size_t len, u8 color)
tty_write_int(struct tty * t, struct devfs_handle * h, const char * buf, size_t size)
ttyaux_write(fs_handle h, char * buf, size_t size, offt * pos)
vfs_write(fs_handle h, void * buf, size_t buf_size)
vfs_writev(fs_handle h, const struct iovec * iov, int iovcnt)
sys_writev(int fd, const struct iovec * u_iov, int u_iovcnt)
do_syscall_int(syscall_type fptr, regs_t * r, _Bool raw_regs)
do_syscall(regs_t * r)
handle_syscall(regs_t * r)
syscall_entry(regs_t * r)
syscall_int80_entry()
```

vvaltchev (changes requested):
OK, I'll merge this and make the bigger changes myself.
Do you mind just fixing the commit message? The tag should be "[term]" instead of "[safe_ringbuf]" because you're fixing the term implementation. Those tags are very useful for searching fixes in a certain area.

PR #190, opened 2024-07-16
[interrupts] Disable irq if do_schedule() just returned in irq_resched()

aenrbes:
Hi @vvaltchev 

In` irq_resched()`,  if `do_schedule()` just returned, then `irq_entry()` will return with interrupts enabled! This doesn't seem to cause any problems with the i386 architecture. But for developers trying to port tilck to other architectures, they often assume that the interrupt is off when `irq_entry()` returns. Therefore, disable interrupts here to avoid causing confusion for developers.

### DIFF
--- a/kernel/interrupts.c
+++ b/kernel/interrupts.c
@@ -188,6 +188,9 @@ static void irq_resched(regs_t *r)
 
       /* Call do_schedule() with preemption disabled, as mandatory */
       do_schedule();
+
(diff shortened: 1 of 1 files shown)


vvaltchev:
Thanks for the fix, @aenrbes. Yes, it makes perfectly sense to make that symmetric.
I guess that was a leftover from the past when do_schedule() never returned. The function just did a full context switch to the same task, if necessary.

PR #194, opened 2024-07-22
Minor coding style fixes in the RISCV64 code

vvaltchev:
@aenrbes I made some minor coding style fixing (formatting, whitespace etc.) to the RISCV64 code. I preferred making those fixes myself to avoid bothering you too many with very "little value" comments.

One change however is non-cosmetic: I moved the `tilck.map` to the build directory `${CMAKE_BINARY_DIR}` because it was written to the root source directory: all the generated files should go in `${CMAKE_BINARY_DIR}` because that could be anywhere. Supporting out-of-tree builds is very important.

I believe this small change shouldn't cause any issues as I didn't see any other references to `tilck.map` but I guess you might have some other code relying on that.

### DIFF
--- a/include/3rd_party/fdt_helper.h
+++ b/include/3rd_party/fdt_helper.h
@@ -1,4 +1,5 @@
 // SPDX-License-Identifier: BSD-2-Clause
+
 /*
  * fdt_helper.h - Flat Device Tree parsing helper routines
  * Implement helper routines to parse FDT nodes on top of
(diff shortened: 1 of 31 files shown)


aenrbes (approved):

aenrbes:
@vvaltchev Thank you so much for doing all the little fixes yourself! I will try my best to fix these problems myself in the next commit.
There is no problem to moved the tilck.map to the build directory.

vvaltchev:
Thank you for the review and for the positive approach, @aenrbes !

PR #195, opened 2024-07-22
[riscv64] Implement exception handling and panic

aenrbes:
Hi @vvaltchev 

In `fdt_serial.c`, I added some temporary code, mainly to enable early serial output. When the serial driver is updated in the future, these temporary code will be overwritten.

### DIFF
--- a/kernel/arch/riscv/debug.c
+++ b/kernel/arch/riscv/debug.c
@@ -13,27 +13,192 @@
 
 #include <elf.h>
 
+static bool fp_is_valid(ulong fp, ulong sp)
+{
+   ulong low = sp + 2 * sizeof(void *);
+   ulong high = (sp + (KERNEL_STACK_SIZE - 1)) & ~(KERNEL_STACK_SIZE - 1);
+
+   if (!fp || fp < low || fp > high || fp & 0x7)
+      return false;
+   else
+      return true;
+}
+
+static size_t
+stackwalk_riscv(void **frames,
+            size_t count,
+            void *fp,
+            pdir_t *pdir)
+{
+   bool curr_pdir = false;
+   void *retAddr;
+   size_t i;
+   ulong old_fp;
+
+   if (!fp) {
+      fp = __builtin_frame_address(0);
+   }
+
+   if (!pdir) {
+      pdir = get_curr_pdir();
+      curr_pdir = true;
+   }
+
+   for (i = 0; i < count; i++) {
+
+      old_fp = (ulong)fp;
+
+      if ((ulong)fp < BASE_VA)
+         break;
+
+      if (curr_pdir) {
+
+         retAddr = *((void **)fp - 1);
+         fp = *((void **)fp - 2);
+
+      } else {
+
+         if (virtual_read(pdir, (void **)fp - 1,
+                          &retAddr, sizeof(retAddr)) < 0)
+            break;
+
+         if (virtual_read(pdir, (void **)fp - 2,
+                          &fp, sizeof(fp)) < 0)
+            break;
+      }
+
+      if (!fp_is_valid((ulong)fp, old_fp)) {
+
+         /*
+          * In cases where the compiler lacks the option
+          * -fno-omit-leaf-frame-pointer, in the leaf function,
+          * fp is saved in the position where ra is in non-leaf case.
+          */
+         if (fp_is_valid((ulong)retAddr, old_fp)) {
+
+            fp = retAddr;
+            continue;
+         }
+
+         break;
+      }
+
+      frames[i] = retAddr;
+   }
+
+   return i;
+}
+
 void dump_stacktrace(void *ebp, pdir_t *pdir)
 {
-   NOT_IMPLEMENTED();
+   void *frames[32] = {0};
+   size_t c = stackwalk32(frames, ARRAY_SIZE(frames), ebp, pdir);
+
+   printk("Stacktrace (%lu frames):\n", c);
+
+   for (size_t i = 0; i < c; i++) {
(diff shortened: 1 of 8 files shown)


kernel/arch/riscv/debug.c
```diff
@@ -13,27 +13,191 @@
...
+
+      old_fp = (ulong)fp;
+
+      if ((ulong)fp < BASE_VA)
+         break;
+
+      if (curr_pdir) {
+
+         retAddr = *((void **)fp - 1);
+         fp = *((void **)fp - 2);
```
vvaltchev, line 57: Purely cosmetic: for symmetry, there should be an empty line before `} else {` since after `} else {` there is an empty line. Also, note that when the `if (...)` line is much shorter than the following statement, if you want you can also skip the empty line and pack things up, but in that case you should also pack the line after `} else {` for symmetry as well. Whatever you prefer.

The reason for the empty line is that a long `if` or `for` followed by a short statement like `void *p;` hides the first short statement visually:
```
if (something(x) && (something_else(x + y) || !special_condition)) {
   void *p;
   do_something_else();
   foo();
}
```

 I always found that hard to read and that's why I started adding those extra lines. But for short statements, that doesn't apply:
```
if (p) {
   call_long_function(a, b, c, d);
}
```
Because the following line stands out a lot.

I realize how complex that might look like to remember. It's simply that I prefer the code to super visually readable instead of being consistent within a simple set of rules. I guess that Tilck's coding style is almost consistent, but the "set of rules" is very non-trivial at least :-)

kernel/arch/riscv/debug.c
```diff
@@ -13,27 +13,191 @@
...
+   ulong high = (sp + (KERNEL_STACK_SIZE - 1)) & ~(KERNEL_STACK_SIZE - 1);
+
+   if (!fp || fp < low || fp > high || fp & 0x7)
+      return false;
+   else
+      return true;
+}
+
+static size_t
+stackwalk32(void **frames,
```
vvaltchev, line 28: On x86, `stackwalk32()` was named like that to recall that's the (x86) 32-bit version of that function. Can you rename this function to something like `stackwalk_riscv()` so that we have less name clashing when searching etc. and so that it's clear that we have a stackwalk_* function for each architecture?

kernel/arch/riscv/fault.c
```diff
@@ -34,33 +34,128 @@ const char *riscv_exception_names[32] =
...
+   }
+
+   if (!cow) {
+
+      if (is_fault_resumable(int_num))
+         return handle_resumable_fault(r);
+
+      if (LIKELY(fault_handlers[int_num] != NULL)) {
+
+         fault_handlers[int_num](r);
```
vvaltchev, line 146: Style (optional): like above, an extra line before `} else {` is needed for symmetry.

kernel/arch/riscv/panic.c
```diff
@@ -31,11 +31,183 @@ regs_t panic_state_regs;
...
+   static va_list saved_args;
+
+   ulong rc;
+   va_list args;
+   struct task *curr;
+
+   if (!kopt_panic_kb) {
+
+      /* No interrupts: we're in a panic state */
+      disable_interrupts_forced();
```
vvaltchev, line 100: Style (optional): same comment as above.

kernel/arch/riscv/panic.c
```diff
@@ -31,11 +31,183 @@ regs_t panic_state_regs;
...
+   panic_save_current_state();
+   curr = get_curr_task();
+
+   if (term_is_initialized()) {
+
+      if (get_curr_tty() != NULL)
+         tty_setup_for_panic(get_curr_tty());
+
+      /* In case the video output has been paused, we MUST restart it */
+      term_restart_output();
```
vvaltchev, line 158: Style (optional): same comment as above.

vvaltchev (changes requested):
Very good change, thanks! 
I have only cosmetic comments plus a request to rename `stackwalk32()`.

kernel/arch/riscv/debug.c
aenrbes: Thank you for explaining it in detail!

kernel/arch/riscv/debug.c
aenrbes: fixed!

kernel/arch/riscv/fault.c
aenrbes: fixed!

kernel/arch/riscv/panic.c
aenrbes: fixed!

kernel/arch/riscv/panic.c
aenrbes: fixed!

kernel/arch/riscv/debug.c
```diff
@@ -13,27 +13,192 @@
...
+   ulong high = (sp + (KERNEL_STACK_SIZE - 1)) & ~(KERNEL_STACK_SIZE - 1);
+
+   if (!fp || fp < low || fp > high || fp & 0x7)
+      return false;
+   else
+      return true;
+}
+
+static size_t
+stackwalk_riscv(void **frames,
```
vvaltchev, line 28: @aenrbes Thanks for the fixes!

Note: here the parameters should be aligned like this:
```
stackwalk_riscv(void **frames,
                size_t count,
                void *fp,
                pdir_t *pdir)
```

But I'm going to merge this PR anyway to unblock you and move forward. You can make this cosmetic fix in your next PR.

PR #196, opened 2024-07-23
[riscv64] Implement virtual memory management

aenrbes:
Hi @vvaltchev

Implement virtual memory management.
Only the three-level page table(sv-39) has been fully tested. (Note: sv-39 is the most commonly used configuration, almost all risc64 chips support sv-39)

### DIFF
--- a/kernel/arch/riscv/paging.c
+++ b/kernel/arch/riscv/paging.c
@@ -33,6 +33,7 @@ ulong kernel_va_pa_offset;
 ulong linear_va_pa_offset;
 
 pdir_t *__kernel_pdir;
+static char kpdir_buf[sizeof(pdir_t)] ALIGNED_AT(PAGE_SIZE);
 
 #define EARLY_PT_NUM 4
 static page_table_t early_pt[EARLY_PT_NUM] ALIGNED_AT(PAGE_SIZE);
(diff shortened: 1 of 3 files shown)


kernel/arch/riscv/debug.c
```diff
@@ -92,7 +92,7 @@ stackwalk_riscv(void **frames,
 void dump_stacktrace(void *ebp, pdir_t *pdir)
 {
    void *frames[32] = {0};
-   size_t c = stackwalk32(frames, ARRAY_SIZE(frames), ebp, pdir);
+   size_t c = stackwalk_riscv(frames, ARRAY_SIZE(frames), ebp, pdir);
```
vvaltchev, line 95: @aenrbes Oops, that means that the previous change broke the RISCV64 build :-(
That's why Tilck has a matrix of CI builds and tests for multiple configurations with multiple compilers. Unfortunately, we don't have one for the RISCV build yet and it will take me some time for me to do that as it's not that simple. The problem with breaking builds even for a single commit is that it prevents `git bisect` from working properly: if the build is broken at that commit, we cannot run a test and check if the regression was introduced at that point or not.

kernel/arch/riscv/debug.c
```diff
@@ -26,9 +26,9 @@ static bool fp_is_valid(ulong fp, ulong sp)
 
 static size_t
 stackwalk_riscv(void **frames,
-            size_t count,
-            void *fp,
-            pdir_t *pdir)
+                size_t count,
```
vvaltchev, line 29: Thanks for that!

kernel/arch/riscv/paging.c
```diff
@@ -47,41 +48,291 @@ static page_table_t *alloc_early_pt(void)
...
+
+   // Allocate a new page.
+   void *new_page_vaddr = kmalloc(PAGE_SIZE);
+
+   if (!new_page_vaddr) {
+
+      // Out-of-memory case
+      struct task *curr = get_curr_task();
+
+      if (!curr->running_in_kernel) {
```
vvaltchev, line 113: In Tilck the preferred pattern is:
```
if (error-check) {
   // nagative case handling
   return error;
}

// positive code path
```

In this case, the `!running_in_kernel` condition is the positive path, when you return true, so that should be left outside of the IF. Also, if possible, no `else` case after a return. In summary, the following will look nicer:

```
if (curr->running_in_kernel) {
   panic();
}

send_signal();
return true;
```

kernel/arch/riscv/paging.c
```diff
@@ -99,23 +352,126 @@ unmap_pages_permissive(pdir_t *pdir,
...
-   NOT_IMPLEMENTED();
+   page_table_t *pt;
+   const ulong vaddr = (ulong) vaddrp;
+
+   ASSERT(!(vaddr & OFFSET_IN_PAGE_MASK)); // the vaddr must be page-aligned
+   ASSERT(!(paddr & OFFSET_IN_PAGE_MASK)); // the paddr must be page-aligned
+
+   for (int level = RV_PAGE_LEVEL; level > 0; level--) {
+
+      pt = PA_TO_LIN_VA(
```
vvaltchev, line 444: Style: if you don't mind:
```
pt = PA_TO_LIN_VA(
   pdir->entries[PTE_INDEX(level, vaddr)].pfn << PAGE_SHIFT
);

File: kernel/arch/riscv/paging.c [hunk 5]
```diff
@@ -126,21 +482,128 @@ map_pages_int(pdir_t *pdir,
...
+   u32 big_page_flags;
+
+   ASSERT(!((ulong)vaddr & OFFSET_IN_PAGE_MASK));
+   ASSERT(!(paddr & OFFSET_IN_PAGE_MASK));
+
+   if (big_pages_allowed && rem_pages >= PTRS_PER_PT) {
+
+      for (; pages < rem_pages; pages++) {
+
+         if (!((ulong)vaddr & (L1_PAGE_SIZE - 1)) &&
```
vvaltchev, line 498: This is hard to read and we'll need in other places to check for the alignment against L1 pages. Can you introduce a nice macro such `PAGING_IS_L1_ALIGNED()` or something like that? You can totally pick a better name, of course.

kernel/arch/riscv/paging.c
```diff
@@ -150,31 +613,255 @@ map_pages(pdir_t *pdir,
...
+   pdir_t *new_pdir = kzalloc_obj(page_table_t);
+
+   if (!new_pdir)
+      return NULL;
+
+   ASSERT(IS_PAGE_ALIGNED(new_pdir));
+   memcpy32(new_pdir, pdir, sizeof(pdir_t) / 4);
+
+   if (pdir_clone_int(pdir, new_pdir, BASE_VADDR_PD_IDX,
+                        RV_PAGE_LEVEL, true))
```
vvaltchev, line 808: Style: this is not aligned.

kernel/arch/riscv/paging.c
```diff
@@ -264,40 +951,268 @@ void *init_early_mapping(ulong fdt_paddr)
...
+   page_table_t *pmd;
+   const ulong vaddr = (ulong) vaddrp;
+   ulong *pte;
+   page_t *e;
+
+   for (int level = RV_PAGE_LEVEL; level > 1; level--) {
+
+      e = &pdir->entries[PTE_INDEX(level, vaddr)];
+
+      ASSERT((e->present) && !(e->raw & _PAGE_LEAF));
```
vvaltchev, line 995: I guess here as well a macro to check for leaf page alignment would help..

kernel/arch/riscv/paging.c
```diff
@@ -264,40 +951,268 @@ void *init_early_mapping(ulong fdt_paddr)
...
+
```
vvaltchev, line 996: This (between the ASSERT and the other statement) is an extra empty line.

kernel/arch/riscv/paging.c
```diff
@@ -264,40 +951,268 @@ void *init_early_mapping(ulong fdt_paddr)
...
+static void
+set_4kb_page_attr(pdir_t *pdir,
+                  void *vaddrp,
+                  ulong attr)
+{
+   page_table_t *pt;
+   ulong *pte;
+   const ulong vaddr = (ulong) vaddrp;
+
+   ASSERT(!(vaddr & OFFSET_IN_PAGE_MASK)); // the vaddr must be page-aligned
```
vvaltchev, line 1015: Yeah, I used OFFSet_IN_PAGE_MASK because (PAGE_SIZE - 1) was too ugly. But a nice set of macros for check for alignment at L0, L1, L2 etc. would nice and could be used of any architecture (with a different definition of course).

kernel/arch/riscv/paging.c
```diff
@@ -264,40 +951,268 @@ void *init_early_mapping(ulong fdt_paddr)
...
+
+   pt = pdir_get_page_table(pdir, vaddr);
+   ASSERT(IS_PAGE_ALIGNED(pt));
```
vvaltchev, line 1018: Here we are. I already had that. So, maybe you can just add `IS_L1_PAGE_ALIGNED()` as it's implicit that PAGE is the smallest one.

kernel/arch/riscv/paging.c
```diff
@@ -264,40 +951,268 @@ void *init_early_mapping(ulong fdt_paddr)
...
-   NOT_IMPLEMENTED();
+   ASSERT(!((ulong)vaddr & OFFSET_IN_PAGE_MASK));
+   ASSERT(IS_PAGE_ALIGNED(size));
+
+   const void *end = vaddr + size;
+
+   while (vaddr < end) {
+
+      if (in_big_page(pdir, vaddr)) {
+
```
vvaltchev, line 1036: Style: extra empty line.

kernel/arch/riscv/paging.c
```diff
@@ -264,40 +951,268 @@ void *init_early_mapping(ulong fdt_paddr)
...
```
vvaltchev, line 1057: Style: extra empty line.

kernel/arch/riscv/paging_generic.c
```diff
@@ -28,32 +28,150 @@ struct kmalloc_heap *hi_vmem_heap;
...
+      return NULL;
+   }
+
+   count = map_kernel_pages(vaddr,
+                            paddr,
+                            page_count,
+                            pg_flags);
+
+   if (count < page_count) {
+
```
vvaltchev, line 53: Style: extra empty line (the following it much longer than the `if` statement).

kernel/arch/riscv/paging_generic.c
```diff
@@ -28,32 +28,150 @@ struct kmalloc_heap *hi_vmem_heap;
...
+   return vaddr + offset;
 }
 
 void iounmap(void *vaddr)
 {
-   NOT_IMPLEMENTED();
+   size_t size = 0;
+
+   ASSERT(IS_PAGE_ALIGNED(vaddr));
+
```
vvaltchev, line 69: Extra line

kernel/arch/riscv/paging_generic.c
```diff
@@ -28,32 +28,150 @@ struct kmalloc_heap *hi_vmem_heap;
...
+   disable_preemption();
+   {
+      per_heap_kfree(hi_vmem_heap, vaddr, &size, 0);
+   }
+   enable_preemption();
+
```
vvaltchev, line 75: Extra line

kernel/arch/riscv/paging_generic.c
```diff
@@ -63,32 +181,130 @@ map_framebuffer(pdir_t *pdir,
                 ulong size,
                 bool user_mmap)
 {
-   NOT_IMPLEMENTED();
+   if (!get_kernel_pdir())
```
vvaltchev, line 184: I think you said that we don't have any framebuffer support on RISCV64 yet. Also, this code mentions WC (write-combining). Do we have anything like that on RISCV ?

vvaltchev (changes requested):
Thanks for following up so quickly with the next change! 
I did only a superficial review looking at the coding style and very obvious things. Tomorrow I'll try to take a deeper look into this code as I'd like to understand it better.

aenrbes:
I'm really sorry, this is really a mistake of mine, is there anything I can do to remedy it now?



---- Replied Message ----
| From | Vladislav K. ***@***.***> |
| Date | 07/23/2024 11:55 |
| To | ***@***.***> |
| Cc | ***@***.***>***@***.***> |
| Subject | Re: [vvaltchev/tilck] [riscv64] Implement virtual memory management (PR #196) |

@vvaltchev commented on this pull request.

In kernel/arch/riscv/debug.c:

> @@ -92,7 +92,7 @@ stackwalk_riscv(void **frames,
 void dump_stacktrace(void *ebp, pdir_t *pdir)
 {
    void *frames[32] = {0};
-   size_t c = stackwalk32(frames, ARRAY_SIZE(frames), ebp, pdir);
+   size_t c = stackwalk_riscv(frames, ARRAY_SIZE(frames), ebp, pdir);


@aenrbes Oops, that means that the previous change broke the RISCV64 build :-(
That's why Tilck has a matrix of CI builds and tests for multiple configurations with multiple compilers. Unfortunately, we don't have one for the RISCV build yet and it will take me some time for me to do that as it's not that simple. The problem with breaking builds even for a single commit is that it prevents git bisect from working properly: if the build is broken at that commit, we cannot run a test and check if the regression was introduced at that point or not.

kernel/arch/riscv/debug.c
aenrbes: I'm really sorry, this is really a mistake of mine, is there anything I can do to remedy it now?

kernel/arch/riscv/paging_generic.c
aenrbes: Yes, there is no support for framebuffer on qemu, but many real-world chips do (such as the popular allwinner D1), I will try to port tilck to real-world chips soon, and framebuffer will be an very important feature at that time;
Some riscv chips (such as allwinner D1, which uses the T-HEAD C906 core) support WC, but it use vendor defined page attribute bits, `early_get_cpu_features()` will set `riscv_cpu_features.page_wt`(vendor defined page attribute bits) according to the specific chip model.

kernel/arch/riscv/paging.c
aenrbes: fixed！

kernel/arch/riscv/paging.c
aenrbes: fixed！

kernel/arch/riscv/paging.c
aenrbes: fixed！

kernel/arch/riscv/paging.c
aenrbes: fixed！

kernel/arch/riscv/paging.c
aenrbes: fixed！

kernel/arch/riscv/paging_generic.c
aenrbes: fixed！

kernel/arch/riscv/paging.c
aenrbes: fixed！

kernel/arch/riscv/debug.c
vvaltchev: @aenrbes I could do a force-push to fix the issue if that won't be a problem for you. I avoid doing that, but if the commit is very recent, it's highly unlikely to upset many people because of that, at least for a project of this scale. For more popular projects that would never be an option.

kernel/arch/riscv/paging_generic.c
aenrbes: fixed！

kernel/arch/riscv/paging_generic.c
aenrbes: fixed！

kernel/arch/riscv/paging.c
aenrbes: fixed!

kernel/arch/riscv/paging.c
aenrbes: fixed!

kernel/arch/riscv/paging.c
aenrbes: fixed!

kernel/arch/riscv/paging.c
aenrbes: fixed!

kernel/arch/riscv/debug.c
aenrbes: I don't have a problem, then please revoke the commit "[riscv64] Implement exception handling and panic", I will submit a new PR immediately

vvaltchev: @aenrbes I wanted to do it myself a very quick fix + `commit --amend` + `push --force` on the previous change, but waited for you to confirm first. There was no need to prepare a PR on the previous commit for a force push. Anyway, it got late in my time zone (Pacific Time) so I disconnected and this morning was already too late for the force-push.

Anyway, no big deal, the RISCV build is still experimental. Just, as the port is getting mature and things start working, we need to be careful to not break it.

Btw, on the full RISCV change did you run `build/st/run_all_tests` ? That's the script runner that runs all the selftests and also the devshell tests.

vvaltchev: @aenrbes Btw, can you rebase this PR on master?

aenrbes: I tried to run `build/st/run_all_tests`, but found that it did not work properly with riscv64, and it may would take a pretty big fix to make `run_all_tests` work, so I manually ran all devshell and selftest commands. In addition, I tested the stability by executing `runall` continuously.

aenrbes: It is worth mentioning that in the `selftest time` test, time drift will be significantly larger than i386,but this is expected and is not a bug; `fork_perf` is significantly slower than the i386, possibly due to three-level page tables, etc

vvaltchev: That's totally OK, thanks for the clarification. The time drift is not a big deal, nor the slower fork_perf. We can optimize that by implement CoW for the page tables. Currently, CoW during fork() is applied only for the regular pages, while all the page tables are just copied.

I'll take a look at that test runner once the full RISCV64 port is completed. Given that on RISCV64 the Tilck runs with the serial port, I guess there won't be a problem to make that runner work, because that's what's used there as well. Just, on i386, for performance and convenience reasons, I ask QEMU to load Tilck instead of Tilck's bootloader. That's also what the ` run_multiboot_qemu` script does (it uses QEMU's `-kernel` option). 

Therefore, for RISCV64 the kernel command line needs to be passed to the U-BOOT bootloader in a different way, maybe by writing to that special config file, I don't know. In the worst case, the test runner can write to serial port any command it wants and run the tests that way.

kernel/arch/riscv/paging_generic.c
vvaltchev: Thanks for clarifying that! I'm so glad that you want to actually make Tilck to work on a real chip, not just on QEMU. That's the approach I took for x86: it must work on real machines, not just on hypervisors. And the challenge there is significant: many things that just work on QEMU don't work on real HW as-it-is due to: extra error cases, race conditions, unexpected system memory layout etc. When you get there, let me know which SoC device you decided to support first, so that I'll buy the same one. I guess it will be something similar to a RPI but for RISCV, right?

Unfortunately, at least for now Tilck cannot work on non-MMU chips, so "true embedded" microcontrollers are out of scope for the moment. But we could get there we if wanted :-)

vvaltchev:
Thanks for adding the macros for the page alignment and fixing the other minor things!
I'd really love to spend more time digging into the implementation details and do a proper review, but I just cannot find the necessary time to do so. Your changes are big, and I don't wanna keep you on hold for a long time until I got the time to understand _everything_. 

Since I don't see any red flags in the code, I'm going to merge this PR and move on. I guess that at least for the moment I'll continue to do quick reviews like that when the code is RISCV-only and I don't have extra time. I'll spend later some time reading your code trying to figure it out all the details. If I find something wrong, I'll write a fix, open a PR, talk with you etc.

kernel/arch/riscv/paging_generic.c
aenrbes: Has the plan to support nommu been put on the agenda?

aenrbes: Are there any concrete plans to support nommu?

vvaltchev: @aenrbes The NOMMU support is a low-priority feature that I planned for the long term after Tilck is ported on architectures useable on embedded chips, like ARM and RISCV. Once that happened, Tilck would be much closer to the "true embedded" world. Therefore, supporting NOMMU would be mostly about the pure-software side of things:

- allow the kernel to work without any virtual memory
- prepare a special build of `busybox` using uclibc-ng which supports NOMMU

In other words, it doesn't look so hard if the kernel works already on the given architecture with MMU. Actually, about part 1, that's why we have "gcc_nocow". In that configuration `fork()` doesn't use the virtual memory and performs a full copy. Over time, we could gradually support in more and more places a NOMMU configuration where the behavior is different. 

Beyond that, after NOMMU is supported, Tilck could start having some real-time features. But all of that is very long term and low priority for now. I just wanted to hear from you what do you think about that? Does it make sense/like the idea or you would go for something completely different?

aenrbes: After porting riscv64 (at least to a real-world riscv64 chip), I plan to take some time to look into the details of nuttx or linux implementations of nommu and, if possible, port those implementations to tilck too.

aenrbes: I have found that there are many very interesting riscv chips (such as espressif's esp32-c3, which has a full wifi protocol stack) that don't have mmu, and I would really like tilck to run on these wireless chips (I know tilck doesn't have network support yet, But maybe lwip can be ported as a temporary solution),

vvaltchev: @aenrbes That would be amazing! I'm so glad to hear that you like so many aspects of this project, including the NOMMU support. 

Just a note to be on the safe side with **copyright**: I'm required to mention that, to avoid copyright violations, Tilck should not have any code close to what exists on Linux or other operating systems. It's OK to read a description about their implementation on articles etc. (e.g. on LWN), but studying closely another implementation by looking at its source code, in particular if it's from a GPL-licensed project, it's definitively a problem and we could be blamed for plagiarism. **I don't think you did that at all**. It's just a note to stay on the safe side. 

If the code is from a compatible license (MIT or BSD 2-clause), in theory we can copy it but then we need to add entry in the NOTICE file. It's still better than getting stuck, if we have no other choice. Btw, that's necessary also for `fdt_helper.c` and `fdt_helper.h` in order to be super on the safe side. 

Even if you completely re-implement the code after studying the other implementation, it could be a problem. It's legal if the other project has a permissive license, but we need to add a note saying that our implementation was heavily inspired by theirs. Because of that we still won't be able to claim "authorship of original work" for all of that code. Only with "public domain" code we're more free, but even there, it's a stretch to claim actual copyright over "public domain" code. While "public domain" comes with no conditions, the implicit condition is that we cannot claim full copyright over that and prevent others from using it. We can however re-distribute "public domain" code under any license we want. 

So.. ideally, the less we take, the better. Unless we do download components in the toolchain. That's OK for many cases because that 3rd-party source is in no way part of the project's source, which should be all result of original work. Still, even in the toolchain, I believe we cannot link GPL projects into Tilck because GPL cannot be re-distributed under a difference license, even in binary form, I think. But we don't distribute binaries for now. So, it's border-line. I hope you got the idea of the kind of problems exist with 3rd-party code.

If you need to take a quick look at some code, seriously avoid Linux and anything GPL.

aenrbes: Thank you for pointing out this problem. I will pay attention to copyright things. In addition, regarding fdt_helper.c and fdt_helper.h, I will add a clarification in NOTICE file immediately.

vvaltchev: @aenrbes Thanks for understanding :-)

PR #197, opened 2024-07-23
Fix a broken commit "[riscv64] Implement exception handling and panic"

aenrbes:
Hi @vvaltchev 

Sorry again for my mistake! 
It was a lesson learned and I will make sure it doesn't happen again in future commit.


vvaltchev:
No problem, @aenrbes. It happens to everyone. I cannot force-push the fix anymore because I checked that 3 forks have been updated in the last 6 yours. So, I'll just submit a fix.

PR #198, opened 2024-07-24
[system_mmap] Fix case where memory is not from zero

aenrbes:
Hi @vvaltchev

`linear_map_mem_region()` previously had an assumption that the memory physical address would start at zero, but that assumption no longer applies.


vvaltchev:
Thanks @aenrbes, it makes sense.

PR #199, opened 2024-07-24
[riscv64] Introduce a simple irqchip framework

aenrbes:
Hi @vvaltchev 

This framework implements concepts similar to linux interrupt domains. The framework resolves the dependencies of interrupt controllers in the device tree and initializes them in order; Peripherals do not need to worry about the interrupt hardware connection when they request an interrupt, the framework will take care of these problems according to the device tree; And the interrupt number is dynamically assigned.
On a typical embedded chip, the number of interrupt controllers is relatively large, and there may be more than three levels of connection between interrupt controllers (such as “intc->plic->gpio->peripherals” on riscv). Therefore, we must use a more flexible approach to management interrupt, rather than using hard-coded interrupt numbers like x86.

### DIFF
--- a/include/tilck/mods/irqchip.h
+++ b/include/tilck/mods/irqchip.h
@@ -1,0 +1,93 @@
+/* SPDX-License-Identifier: BSD-2-Clause */
+
+#pragma once
+
+#include <tilck/common/basic_defs.h>
+#include <tilck/kernel/list.h>
+
+#define FDT_MAX_INTERRUPTS_PARAMS 16
+
+enum irq_type {
+   IRQ_TYPE_NONE           = 0x00000000,
+   IRQ_TYPE_EDGE_RISING    = 0x00000001,
+   IRQ_TYPE_EDGE_FALLING   = 0x00000002,
+   IRQ_TYPE_EDGE_BOTH      = (IRQ_TYPE_EDGE_FALLING | IRQ_TYPE_EDGE_RISING),
+   IRQ_TYPE_LEVEL_HIGH     = 0x00000004,
+   IRQ_TYPE_LEVEL_LOW      = 0x00000008,
+};
+
+struct fdt_irq_param {
+   int intc_node;
+   int param_nums;
+   u32 params[FDT_MAX_INTERRUPTS_PARAMS];
+};
+
+struct fdt_irqchip {
+   struct list_node node;
+   const struct fdt_match *id_table;
+   int (*init)(void *fdt, int node, const struct fdt_match *match);
+};
+
+struct fdt_irqchip_ops {
+   void (*hwirq_set_mask)(int hwirq, void *priv);
+   void (*hwirq_clear_mask)(int hwirq, void *priv);
+   bool (*hwirq_is_masked)(int hwirq, void *priv);
+   void (*hwirq_set_type)(int hwirq, int type, void *priv);
+   int (*xlate_irq_param)(struct fdt_irq_param *param,
+                          int *hwirq, int *type, void *priv);
+};
+
+struct irq_domain {
+   void *priv;
+   struct list_node node;
+   int fdt_node;
+   struct fdt_irqchip_ops *ops;
+   size_t irq_map_size;
+   int irq_map[];
+};
+
+struct irq_data {
+   bool present;
+   u32 hwirq;
+   enum irq_type type;
+   u32 unhandled_count;
+   struct irq_domain *domain;
+};
+
+#define REGISTER_FDT_IRQCHIP(__name, __id_table, __init)  \
+                                                          \
+   static struct fdt_irqchip fdt_irqchip_##__name = {     \
+      .id_table = __id_table,                             \
+      .init = __init                                      \
+   };                                                     \
+                                                          \
+   __attribute__((constructor))                           \
+   static void __register_fdt_irqchip_##__name(void) {    \
+      irqchip_drv_register(&fdt_irqchip_##__name);        \
+   }
+
+int fdt_parse_one_hwirq(void *fdt,
+                        int node,
+                        int index,
+                        struct fdt_irq_param *irq_param);
+
+int default_xlate_irq_param(struct fdt_irq_param *param,
+                            int *hwirq,
+                            int *type,
+                            void *priv);
+
+struct irq_domain *
+irqchip_register_irq_domain(void *priv,
+                            int fdt_node,
+                            int int_nums,
+                            struct fdt_irqchip_ops *ops);
+
+void irqchip_drv_register(struct fdt_irqchip *drv);
+int irqchip_get_irq_count(void *fdt, int node);
+int irqchip_alloc_irq(void *fdt, int node, int hwirq_index);
+void irqchip_free_irq(int irq);
+int irqchip_get_free_irq(struct irq_domain *domain, int hwirq);
+void irqchip_put_irq(int irq);
+enum irq_action generic_irq_handler(u8 irq);
+void init_fdt_irqchip(void);
+
--- a/kernel/arch/riscv/irq.c
+++ b/kernel/arch/riscv/irq.c
@@ -12,48 +12,124 @@
 #include <tilck/kernel/worker_thread.h>
 #include <tilck/kernel/timer.h>
 
+#include <tilck/mods/irqchip.h>
+
+extern struct irq_data irq_datas[MAX_IRQ_NUM];
+extern ulong irq_bitmap[MAX_IRQ_NUM / (sizeof(ulong) * 8)];
 
 struct list irq_handlers_lists[MAX_IRQ_NUM];
 
 void irq_set_mask(int irq)
 {
-   NOT_IMPLEMENTED();
+   ulong var;
+   u32 hwirq;
+   struct irq_domain *domain;
+   ASSERT(IN_RANGE_INC(irq, 0, MAX_IRQ_NUM - 1));
+
+   disable_interrupts(&var);
+   if (irq_datas[irq].present) {
+      domain = irq_datas[irq].domain;
+      hwirq = irq_datas[irq].hwirq;
+      domain->ops->hwirq_set_mask(hwirq, domain->priv);
+   }
+   enable_interrupts(&var);
 }
 
 void irq_clear_mask(int irq)
 {
-   NOT_IMPLEMENTED();
+   ulong var;
+   u32 hwirq;
+   struct irq_domain *domain;
+   ASSERT(IN_RANGE_INC(irq, 0, MAX_IRQ_NUM - 1));
+
+   disable_interrupts(&var);
+   if (irq_datas[irq].present) {
+      domain = irq_datas[irq].domain;
+      hwirq = irq_datas[irq].hwirq;
+      domain->ops->hwirq_clear_mask(hwirq, domain->priv);
+   }
+   enable_interrupts(&var);
 }
 
 bool irq_is_masked(int irq)
 {
-   NOT_IMPLEMENTED();
+   ulong var;
+   u32 hwirq;
+   struct irq_domain *domain;
+   bool ret;
+
+   ASSERT(IN_RANGE_INC(irq, 0, MAX_IRQ_NUM - 1));
+
+   disable_interrupts(&var);
+   if (irq_datas[irq].present) {
+      domain = irq_datas[irq].domain;
+      hwirq = irq_datas[irq].hwirq;
+      ret = domain->ops->hwirq_is_masked(hwirq, domain->priv);
+   }
+   enable_interrupts(&var);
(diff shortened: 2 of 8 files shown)


kernel/arch/riscv/irq.c
```diff
@@ -12,48 +12,127 @@
...
 {
-   NOT_IMPLEMENTED();
+   ulong var;
+   u32 hwirq;
+   struct irq_domain *domain;
+   ASSERT(IN_RANGE_INC(irq, 0, MAX_IRQ_NUM - 1));
+
+   disable_interrupts(&var);
+   if (irq_datas[irq].present) {
+
```
vvaltchev, line 31: Extra empty line because the following line is longer than the '{' on the IF line.

kernel/arch/riscv/irq.c
```diff
@@ -12,48 +12,127 @@
...
```
vvaltchev, line 48: Extra empty line because the following line is longer than the '{' on the IF line.

kernel/arch/riscv/irq.c
```diff
@@ -12,48 +12,127 @@
...
+   ulong var;
+   u32 hwirq;
+   struct irq_domain *domain;
+   bool ret;
+
+   ASSERT(IN_RANGE_INC(irq, 0, MAX_IRQ_NUM - 1));
+
+   disable_interrupts(&var);
+   if (irq_datas[irq].present) {
+
```
vvaltchev, line 67: Extra empty line because the following line is longer than the '{' on the IF line.

kernel/arch/riscv/mmap.c
```diff
@@ -5,18 +5,23 @@
...
-#define MAX_DMA                                (256 * KB)
-
-STATIC_ASSERT(MAX_DMA <= 16 * MB);
-STATIC_ASSERT((MAX_DMA & (64 * KB - 1)) == 0);
-
 void arch_add_initial_mem_regions()
 {
-   NOT_IMPLEMENTED();
+   /*
+    * Reserve 1MB below kernel's head, where
```
vvaltchev, line 11: Exactly! Since we have a region below Tilck's kernel, why we need a section too?
Also, we have `.data` and `.bss`. Why we need also `.sdata` and `.sbss` ?

modules/irqchip/riscv/fdt_irqchip.c
```diff
@@ -11,6 +11,358 @@
...
+   rc = fdt_parse_one_hwirq(fdt, node, hwirq_index, &irq_param);
+   if (rc)
+      return 0;
+
+   disable_preemption();
+
+   /* Find the corresponding irq domain */
+   list_for_each(domain, tmp, &irq_domains, node) {
+
+      if (domain->fdt_node == irq_param.intc_node) {
```
vvaltchev, line 208: Can you do:
```
if (domain->fdt_node != irq_param.intc_node)
   continue;
```

Instead? That will one extra level of indentation will go away.

vvaltchev: In alternative, you can do the `if` in the positive case, but extract the body of that `if` to a separate function, so that it's easier to read.

modules/irqchip/riscv/fdt_irqchip.c
```diff
@@ -11,6 +11,358 @@
...
+
+   domain->fdt_node = fdt_node;
+   domain->irq_map_size = int_nums;
+   domain->ops = ops;
+   domain->priv = priv;
+
+   disable_preemption();
+   list_add_tail(&irq_domains, &domain->node);
+   enable_preemption();
+
```
vvaltchev, line 284: Extra empty line before the `return` statement.

modules/irqchip/riscv/riscv-plic.c
```diff
@@ -10,8 +10,280 @@
...
+}
+
+static const struct fdt_match plic_ids[] = {
+   {.compatible = "riscv,plic0"},
+   {.compatible = "sifive,plic-1.0.0"},
+   {.compatible = "thead,c900-plic"},
+   { }
+};
+
+DEFINE_FDT_IRQCHIP(plic) = {
```
vvaltchev, line 285: That's very nice!

include/tilck/mods/irqchip.h
```diff
@@ -0,0 +1,101 @@
...
+                     struct fdt_irqchip fdt_irqchip_##_name   \
+                                      __attribute__((used))   \
+        __attribute__((section (".modules.fdt_irqchip.2")))
+
+#define FDT_IRQCHIP_START                                     \
+({                                                            \
+                         static struct fdt_irqchip start[0]   \
+                          __attribute__((aligned(NBITS/8)))   \
+                                      __attribute__((used))   \
+        __attribute__((section (".modules.fdt_irqchip.1")));  \
```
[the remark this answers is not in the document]
aenrbes: First of all, explain`.simplebus_driver`, this is supposed to be useless code that I forgot to clean up, and I will delete it immediately.
For`.modules.*`, I have to make a clarification here. First, these segments have nothing to do with the bootloader or fdt. You can observe the `struct fdt_irqchip`, which actually represents the **driver** of an interrupt controller (which records the address of the driver initialization function and the name of the matching device), rather than an abstraction of the interrupt controller **device**;
The interrupt controllers **driver** register themselves with `DEFINE_FDT_IRQCHIP`, and then we combine these `struct fdt_irqchip` into an array of `.modules.fdt_irqchi.2` via the link script; In this way, `init_fdt_irqchip()` can look up the interrupt controller nodes in the device tree in turn, and then iterate through the `.modules.fdt_irqchi.2` array for each node to see if it can find a matching **driver**, and call the driver initialization function if there is one. So `init_fdt_irqchip()` needs to use `FDT_IRQCHIP_START` to know where the `.modules.fdt_irqchi.2` array starts and ends, This cannot be done using tilck's existing modules initialization hooks alone.

kernel/arch/riscv/mmap.c
aenrbes: The`.sdata` section does not have any effect at the moment, because `-Wl,--no-relax` is now used, this option will be removed in the future, which means that the linker will automatically move some commonly accessed global variables to the `.sdata` section. Then use the `gp`(global pointer) register (will always pointing right in the middle of `.sdata`) to access them, which speeds things up.

include/tilck/mods/irqchip.h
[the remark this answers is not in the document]
vvaltchev: @aenrbes If that has nothing to do with the bootloader, then it's much easier. You can mark as many functions as you need with `__attribute__((constructor))`: they will be automatically called during the early initialization in kmain(). In those functions, each driver can append a struct of its own on a global linked list. You really don't need any sort of fancy extra section for things like that. Look at the logic for selftests:

```
struct self_test {
   struct list_node node;
   const char *name;
   enum se_kind kind;
   void (*func)(void);
};

#define REGISTER_SELF_TEST(__name, __kind, __func)  \
                                                                \
   static struct self_test se_##__name##_inst = {               \
      .name = #__name,                                          \
      .kind = __kind,                                           \
      .func = __func                                            \
   };                                                           \
                                                                \
   __attribute__((constructor))                                 \
   static void __register_se_##__name(void) {                   \
      se_register(&se_##__name##_inst);                         \
   }
```

`se_register()` is just:
```
void se_register(struct self_test *se)
{
   list_add_tail(&se_list, &se->node);
}
```


The REGISTER_SELF_TEST() macro used like this:
```
REGISTER_SELF_TEST(kmalloc_perf, se_long, &selftest_kmalloc_perf)
```

**Note:** you probably don't even need all the macro stuff here. You just need to a constructor function that simply appends an "ops" struct to a linked list.

Once you have a linked list with structs, you can probe all the drivers, one by one. It's flexible because it doesn't require a single place where all the drivers are declared upfront and it doesn't use any other custom section other than the `ctors` one, that is a well-known section used by compilers.

Isn't your problem **completely solvable** with this approach? If not, why?

kernel/arch/riscv/mmap.c
vvaltchev: OK, I did some quick research. Correct me if I'm wrong: RISC-V has the concept of a "global pointer" (GP) which can be configured in the CPU and then we use have GP-relative addressing. That reduces the code size compared to a GP-less mode. That's cool!

So, RISC-V doesn't have anything like RIP-relative addressing, right? 
Anyway, got it. The "s*" sections are used for the data accessing using that GP-relative mode. OK, I'm totally cool with that.

Thanks!

include/tilck/mods/irqchip.h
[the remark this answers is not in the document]
aenrbes: Thank you for pointing it out and I will change it according to this approach.

kernel/arch/riscv/mmap.c
aenrbes: As I understand it, riscv's gcc option `-mcmodel =medany` is basically equivalent to RIP-relative addressing on x86

include/tilck/mods/irqchip.h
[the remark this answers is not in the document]
vvaltchev: Thank you for understanding and working on a fix.

include/tilck/mods/irqchip.h
```diff
@@ -0,0 +1,101 @@
...
+#define FDT_MAX_INTERRUPTS_PARAMS 16
+   u32 params[FDT_MAX_INTERRUPTS_PARAMS];
+};
+
+struct fdt_irqchip {
+   const struct fdt_match *id_table;
+   int (*init)(void *fdt, int node, const struct fdt_match *match);
+};
+
+struct fdt_irqchip_ops {
```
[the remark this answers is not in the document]
aenrbes: In theory, it is possible to refactor the x86 interrupt architecture using some of the infrastructure of fdt_irqchip, but this is a slightly larger change, I can try it after riscv64 porting is complete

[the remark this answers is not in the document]
vvaltchev: Yeah, agreed. I'm not asking for that now. It's OK for the moment to have a divergence and fix that later. Not a big deal at all.

include/tilck/mods/irqchip.h
```diff
@@ -0,0 +1,101 @@
...
+   IRQ_TYPE_EDGE_FALLING   = 0x00000002,
+   IRQ_TYPE_EDGE_BOTH      = (IRQ_TYPE_EDGE_FALLING | IRQ_TYPE_EDGE_RISING),
+   IRQ_TYPE_LEVEL_HIGH     = 0x00000004,
+   IRQ_TYPE_LEVEL_LOW      = 0x00000008,
+};
+
+struct fdt_irq_param {
+   int intc_node;
+   int param_nums;
+#define FDT_MAX_INTERRUPTS_PARAMS 16
```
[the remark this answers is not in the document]
aenrbes: fixed！

kernel/arch/riscv/irq.c
aenrbes: fixed！

kernel/arch/riscv/irq.c
aenrbes: fixed！

kernel/arch/riscv/irq.c
aenrbes: fixed！

modules/irqchip/riscv/fdt_irqchip.c
aenrbes: That makes sense！

modules/irqchip/riscv/fdt_irqchip.c
aenrbes: fixed！

kernel/arch/riscv/mmap.c
vvaltchev: Aha, that’s interesting; we can play with multiple models.

kernel/arch/riscv64/linker_script.ld
```diff
@@ -48,20 +48,6 @@ SECTIONS
       *(.init_array)
    } : ro_segment
 
-   . = ALIGN(4);
-   .modules : AT(kernel_text_paddr + (modules - text))
```
vvaltchev, line 52: Thank you!

kernel/arch/riscv64/linker_script.ld
```diff
@@ -86,12 +72,6 @@ SECTIONS
       bss = .;
       *(.bss .bss.* .gnu.linkonce.b.*)
    } : rw_segment
-
-   .sbss : AT(kernel_text_paddr + (sbss - text))
```
vvaltchev, line 90: Just for curiosity: why removing sbss but keeping sdata?

PR #201, opened 2024-07-25
[riscv64] Implement process management on riscv

aenrbes:
Hi @vvaltchev 

Please review, thanks:-)

### DIFF
--- a/include/tilck/kernel/hal.h
+++ b/include/tilck/kernel/hal.h
@@ -7,7 +7,7 @@
 #include <tilck/common/datetime.h>
 #include <tilck/kernel/hal_types.h>
 
-#define USERMODE_STACK_ALIGN              16u
+#define USERMODE_STACK_ALIGN              16ul
 
 #if defined(__i386__) || defined(__x86_64__)
 
--- a/kernel/arch/riscv/process_riscv.c
+++ b/kernel/arch/riscv/process_riscv.c
@@ -38,16 +38,6 @@ STATIC_ASSERT(
 
 STATIC_ASSERT(sizeof(struct task_and_process) <= 2048);
 
-void setup_pause_trampoline(regs_t *r)
-{
-   NOT_IMPLEMENTED();
-}
-
-void restore_regs_from_user_stack(regs_t *r)
-{
-   NOT_IMPLEMENTED();
-}
-
 int
 push_args_on_user_stack(regs_t *r,
                         const char *const *argv,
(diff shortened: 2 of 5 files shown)


vvaltchev:
Nice change! But that will take some time for me to review, as I should totally spend some time into the details.

I suspect from that now that after this is merged I should spend some effort trying to generalize that code and reduce the arch-specific stuff. I couldn't do that properly before because Tilck didn't run on anything else other than i686. So this is amazing :-)

kernel/arch/riscv/process.c
```diff
@@ -96,12 +445,110 @@ int setup_process(struct elf_program_info *pinfo,
                   struct task **ti_ref,
                   regs_t *r)
 {
-   NOT_IMPLEMENTED();
```
vvaltchev, line 99: I moved this function alone with others to the portable `process.c` code, to help you with this port.

kernel/arch/riscv/process.c
```diff
@@ -64,29 +221,221 @@ int setup_sig_handler(struct task *ti,
...
+      if (sig_state == sig_pre_syscall)
+         r->a0 = (ulong) -EINTR;
+
+      if ((rc = save_regs_on_user_stack(ti, r)) < 0)
+         return rc;
+   }
+
+   r->sepc = user_func;
+   r->usersp -= SIG_HANDLER_ALIGN_ADJUST;
+   r->a0 = (ulong)signum;
```
vvaltchev, line 237: Can you move this code to signal.c and make it portable using the inline funcs for accessing `usersp`, `ip` and other typical register meta-names that are schared across architectures ?

vvaltchev (changes requested):
Hi @aenrbes,
After looking at your port, I saw that many functions could be easily enough made portable. So I wrote PR https://github.com/vvaltchev/tilck/pull/202. If nothing breaks, I'll merge it ASAP.

I still left some functions that can be ported with some effort and didn't review at all the context switching code at the moment. Let's do it as a multi-step review process, after you rebase your PR on the top of my refactoring. The change should be much easier to review. Maybe even some of the context switching code can be made portable.

Thanks,
Vlad

vvaltchev:
@aenrbes Btw, do you mind fixing the commit message and PR message? There's a typo: "pcrocess" => "process".

kernel/arch/riscv/process.c
```diff
@@ -40,13 +40,170 @@ STATIC_ASSERT(sizeof(struct task_and_process) <= 2048);
 
 void task_info_reset_kernel_stack(struct task *ti)
 {
+   ulong bottom = (ulong)ti->kernel_stack + KERNEL_STACK_SIZE;
```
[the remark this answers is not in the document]
aenrbes: This is supposed to be an oversight on my part, and it didn't cause a problem before because I put a padding in the last place of regs_t.

kernel/arch/riscv/process.c
```diff
@@ -40,13 +40,170 @@ STATIC_ASSERT(sizeof(struct task_and_process) <= 2048);
...
-   NOT_IMPLEMENTED();
+   if (rem > 0) {
+      char smallbuf[16] = {0};
+      memcpy(&smallbuf, str + aligned_len, rem);
+      memcpy((void *)(*usersp + aligned_len), &smallbuf, sizeof(smallbuf));
+   }
+}
+
+static int
+push_args_on_user_stack(ulong *const usersp,
```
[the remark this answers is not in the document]
aenrbes: fixed.

kernel/arch/riscv/process.c
aenrbes: fixed.

kernel/user.c
```diff
@@ -276,8 +276,9 @@ void push_string_on_user_stack(regs_t *r, const char *str)
    memcpy(TO_PTR(user_sp), str, aligned_len);
 
    if (rem > 0) {
-      ulong smallbuf = 0;
+      char smallbuf[USER_STACK_STRING_ALIGN] = {0};
```
aenrbes, line 279: Here's an fix

include/tilck/kernel/hal.h
```diff
@@ -7,7 +7,7 @@
 #include <tilck/common/datetime.h>
 #include <tilck/kernel/hal_types.h>
 
-#define USERMODE_STACK_ALIGN              16u
+#define USERMODE_STACK_ALIGN              16ul
```
aenrbes, line 10: fixed here

kernel/user.c
vvaltchev: Good catch, thanks!

kernel/arch/riscv/process_riscv.c
```diff
@@ -91,57 +291,337 @@ void save_current_task_state(regs_t *r,  bool irq)
...
+   if (!is_kernel_thread(curr) && curr->state != TASK_STATE_ZOMBIE)
+      save_curr_fpu_ctx_if_enabled();
+
+   if (!is_kernel_thread(ti)) {
+
+      if (get_curr_pdir() != ti->pi->pdir) {
+         set_curr_pdir(ti->pi->pdir);
+      }
+
+      if (!ti->running_in_kernel && !(state->sstatus & SR_SPP))
```
aenrbes, line 479: I found here just check` ! ti->running_in_kernel` is not enough; On riscv, only `! (state->sstatus & SR_SPP)` indicates that you are about to switch to user space. I think the same problem may need to be fixed on the x86 architecture.

vvaltchev: @aenrbes Mmm.. that's interesting. Do you mind to elaborate a bit more? Is SR_SPP (Supervisor Previous Privilege) set by the CPU ? I have no idea how this works on RISCV.

On x86 we have IOPL should do the job because when switching to usermode IOPL must be 3. However, can you help me understand how `ti->running_in_kernel` can be false while we're going to switch to kernel mode? IRQs are non-preemptable and the syscall handling entry path disables the preemption before re-enabling the interrupts (see syscall_entry). Therefore, at least with a syscall I don't see how this could happen. (But that doesn't mean that I'm not missing anything!)

In case of anything else like an IRQ or a page fault, the `running_in_kernel` flag should not be set anyways..

vvaltchev (approved):
Thanks for updating your PR! However, there is still much to refactor and share more code with x86. I guess it will be faster for me to just merge now, and refactor myself as a follow-up change. Also note that I added a riscv64 build job in the CI/CD builds, so that now we can catch at least trivial build errors.

kernel/arch/riscv/process_riscv.c
aenrbes: @vvaltchev 
When COW occurs from user space and `fault_entry()` is entered, `running_in_kernel` is not set at the time, and then when `enable_preemption()` in `fault_entry()` is executed, there may be a task switch. When it switch back here, `running_in_kernel` is zero but it returns to the kernel context.

vvaltchev: @aenrbes Ouch, that's looks like a problem. Can't we somehow call `set_current_task_in_kernel()` in `fault_entry()` to solve that?

```C
void fault_entry(regs_t *r)
{
   /*
    * Here preemption could be either enabled or disabled. Typically is enabled,
    * but it's totally possible for example a page fault to occur in the kernel
    * while preemption is disabled.
    */
   ASSERT(!are_interrupts_enabled());

   set_current_task_in_kernel();                // <<<<<<<<<< set running_in_kernel here
   push_nested_interrupt(regs_intnum(r));
   disable_preemption();
   enable_interrupts_forced();

   handle_fault(r);

   /*
    * Pop the nested interrupt and process signals while preemption is
    * still disabled.
    */
   pop_nested_interrupt();

   if (!get_curr_task()->running_in_kernel)
      process_signals(get_curr_task(), sig_in_fault, r);

   enable_preemption();
   set_current_task_in_user_mode();   //  <<<<<<<<<<<<<< disable it here
   disable_interrupts_forced();
}
```

**HOWEVER** , in case of a **nested fault** that would not work correctly. But if we make `running_in_kernel` to be an integer instead of a bool and do increase/decrease ? 

Not sure about the `irq_resched()` idea you mentioned. Even if we don't enable the preemption here, inside the fault handler the preemption can still be enabled. That's actually a pretty important use-case, and it's regular to get blocked and sleep in fault handlers in other OSs in order to do I/O and what not. 

Other OSs however don't allow page faults in the kernel due to CoW, but do `page_get()` and `page_put()` all the time. I was wondering if that can be avoided in Tilck, given the simplification that swap won't be supported. So if PF occurs, the page either was uncommitted or it's a CoW and requires a copy. In theory, this could work also with memory-mapped disk files, as long as the pages as always present. The I/O writes can be flushed to disk in an async fashion as long as the initial read happens on the first map. If I'm missing something else instead, we'd need to handle yet another case of PF where actual I/O is done. I was really trying hard to avoid all of that complexity in Tilck to not reimplement 100% of a classic *NIX kernel. I'm trying to keep it as simple as possible and yet.. it's never that simple once we have CoW, over-commitment, signals and what not :-( 

Anyway, do you think the idea of a ref-counted "in kernel" variable would work?

aenrbes: I think a ref-counted "in kernel" variable is still problematic, because there are mechanisms in tilck which assume that if `running_in_kernel` is true, then it is represented in syscall, such as the one in `handle_potential_cow()`.

aenrbes: The main problem is that the new "in kernel" variable needs to be able to distinguish between syscall and fault

vvaltchev: @aenrbes Mmm.. can you elaborate more on that? Tilck does not have cases where a page fault in the kernel is not triggered by a userspace process. So, if we get a page fault in handle_potential_cow(), it's OK to kill the app even if the page fault happened directly, instead of going through a syscall. Actually now it will work only in that case. If the PF happened in kernel while running a syscall, we'll panic the machine. 

Anyway, I'll think about that, but if you have a good idea in the meanwhile, please propose a fix.

PR #203, opened 2024-07-28
Share more code between RISCV64 and x86

vvaltchev:
Hi @aenrbes,
In this PR I made more refactoring to share code between RISCV64 and x86. Two functions remain still because I'd prefer you to do that in a small change (I already merged your previous one):

### setup_sig_handler()

This is very similar between the two architectures, except for:
```
r->a0 = (ulong) -EINTR;
```
Which should be trivially fixable and this assert:
```
ASSERT((regs_get_usersp(r) & (USERMODE_STACK_ALIGN - 1)) == 0);
```

Which is different than the x86 one. Can you please check my comment in the x86 code and see how it applies to RISCV64 and if not, why? I didn't want to try sharing that code because I cannot test RISCV64 really at the moment. But yeah, the shared implementation should go in `signal.c`.

### kthread_create2()

This is similar too, indeed. The regs part:
```
   regs_t r =  {
      .kernel_resume_pc = (ulong)&asm_trap_entry_resume,
      .sepc = (ulong)func,
      .sstatus = SR_SPIE | SR_SPP | SR_SIE | SR_SUM,
   };
```
Can be made to work with #ifdefs and placed in process.c. This part:
```
   r.a0 = (ulong)arg;
   r.ra = (ulong)&kthread_exit;
   r.sp = (ulong)ti->state_regs;
```

Can be made to work with the proper inline functions for accessing registers using their meta-names, as we did in other places. This should go in `process.c`. Do you mind making this small change? Thank you!


aenrbes:
> Hi @aenrbes, In this PR I made more refactoring to share code between RISCV64 and x86. Two functions remain still because I'd prefer you to do that in a small change (I already merged your previous one):
> 
> ### setup_sig_handler()
> This is very similar between the two architectures, except for:
> 
> ```
> r->a0 = (ulong) -EINTR;
> ```
> 

fixed.

> Which should be trivially fixable and this assert:
> 
> ```
> ASSERT((regs_get_usersp(r) & (USERMODE_STACK_ALIGN - 1)) == 0);
> ```
> 
> Which is different than the x86 one. Can you please check my comment in the x86 code and see how it applies to RISCV64 and if not, why? I didn't want to try sharing that code because I cannot test RISCV64 really at the moment. But yeah, the shared implementation should go in `signal.c`.
> 

When riscv's call instruction is executing, the return address is stored in the `ra` register without a stack push.

> ### kthread_create2()
> This is similar too, indeed. The regs part:
> 
> ```
>    regs_t r =  {
>       .kernel_resume_pc = (ulong)&asm_trap_entry_resume,
>       .sepc = (ulong)func,
>       .sstatus = SR_SPIE | SR_SPP | SR_SIE | SR_SUM,
>    };
> ```
> 
> Can be made to work with #ifdefs and placed in process.c. This part:
> 
> ```
>    r.a0 = (ulong)arg;
>    r.ra = (ulong)&kthread_exit;
>    r.sp = (ulong)ti->state_regs;
> ```
> 
> Can be made to work with the proper inline functions for accessing registers using their meta-names, as we did in other places. This should go in `process.c`. Do you mind making this small change? Thank you!

Fixed.

PR #205, opened 2024-07-29
[process][riscv] Share kthread_create2() with RISCV64

vvaltchev:


vvaltchev:
Hi @aenrbes, do you mind taking a look at this PR?
It should be simple, but I wrote it very quickly. Might have missed something.

aenrbes (approved):

PR #211, opened 2024-07-29
[riscv64] Minimal support for riscv64 system calls

aenrbes:
Hi @vvaltchev 

This is milestone for the riscv64 port, now it can successfully complete all system tests.


vvaltchev:
Hi @aenrbes, thanks for making this change.
It looks good overall, even if there are quite some refactoring opportunities and several cosmetic things to fix.
Since I have a bit of time today, I think it's worth merging this PR as-it-is and refactoring it myself with follow-up changes.

PR #212, opened 2024-07-30
[sched] Improve running_in_kernel

aenrbes:
Hi @vvaltchev 

As we discussed in #201, here is a proposal for improve the running_in_kernel.

Make `running_in_kernel` a bit-field variable, with the highest bit indicating weather we are in a syscall and low bits for fault count.

### DIFF
--- a/include/tilck/kernel/sched.h
+++ b/include/tilck/kernel/sched.h
@@ -13,6 +13,8 @@
 
 #include <tilck_gen_headers/config_sched.h>
 
+#define IN_SYSCALL_FLAG (1u << 31)
+
 enum task_state {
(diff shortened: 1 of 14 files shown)


include/tilck/kernel/sched.h
```diff
@@ -63,7 +65,7 @@ struct task {
    struct process *pi;
 
    bool is_main_thread;                      /* value of `tid == pi->pid` */
-   bool running_in_kernel;
+   u32 running_in_kernel;
```
vvaltchev, line 68: Can you move this u32 above `is_main_thread` so that the booleans are packed?

vvaltchev (changes requested):

PR #213, opened 2024-07-31
[kmalloc] Check if internal_kmalloc() returns NULL

aenrbes:
@vvaltchev 

Here seem to have forgotten to consider the case where `internal_kmalloc()` returns NULL.
If `internal_kmalloc()` returns NULL, just return NULL in per_heap_kmalloc_unsafe().

```
   void *big_block =
      internal_kmalloc(h, rounded_up_size, 0, h->size, false, false);

   if (!big_block)    //<<<<<<<<<<<<<<<<<<<<<here takes into account the case where internal_kmalloc returns NULL
      return NULL;

   const int big_block_node = ptr_to_node(h, big_block, rounded_up_size);
   size_t tot = 0;

   for (int i=(int)h->heap_data_size_log2-1; i >= 0 && tot < desired_size; i--)
   {
      const size_t s = (1u << i);

      if (!(desired_size & s))
         continue;

      addr = internal_kmalloc(h,
                              s,
                              big_block_node,
                              rounded_up_size,
                              true,              /* mark node as allocated */
                              do_actual_alloc);

      ASSERT(addr == big_block + tot);       //<<<<<<<<<<<<<<<<<<<<<if internal_kmalloc() just returns NULL, the ASSERT will be triggered！
```

### DIFF
--- a/kernel/kmalloc/kmalloc.c
+++ b/kernel/kmalloc/kmalloc.c
@@ -503,6 +503,9 @@ per_heap_kmalloc_unsafe(struct kmalloc_heap *h, size_t *size, u32 flags)
                               true,              /* mark node as allocated */
                               do_actual_alloc);
 
+      if (!addr)
+         return NULL;
+
       ASSERT(addr == big_block + tot);
 
       if (sub_blocks_min_size) {


kernel/kmalloc/kmalloc.c
```diff
@@ -503,6 +503,9 @@ per_heap_kmalloc_unsafe(struct kmalloc_heap *h, size_t *size, u32 flags)
                               true,              /* mark node as allocated */
                               do_actual_alloc);
 
+      if (!addr)
```
vvaltchev, line 506: For linearly mapped heaps, because we already checked with the `big_block` above that we have that memory "on paper" and we don't need to do anything else, that call is guaranteed to succeed (`do_actual_alloc` is ignored).

However, in the general case, `actual_allocate_node()` might fail, so I guess we should handle that case, if I'm not forgetting something else. I wrote that code a long time ago.

The problem is that, to handle that case correctly you'd need to carefully do a **full roll-back** of all the previous allocations in the loop that succeeded, otherwise we'll have a memory leak.

vvaltchev (changes requested):
Thanks for noticing this, @aenrbes!

This is a nice catch, but I don't think the corner case is handled correctly at the moment, as explained in the comment.

kernel/kmalloc/kmalloc.c
aenrbes: It did fail in `actual_allocate_node()`. more specifically, the fail happened in
```
h>valloc_and_map(alloc_block_vaddr, h->alloc_block_size / PAGE_SIZE) => 
user_map_zero_page() => 
map_zero_pages() => 
map_zero_page() => 
map_page_int() =>        
   pt = kzalloc_obj(page_table_t);
   if (UNLIKELY(! pt))
      return -ENOMEM   //<<<<<<<<<<<<This may occur when testing fork_oom
```

aenrbes: > The problem is that, to handle that case correctly you'd need to carefully do a **full roll-back** of all the previous allocations in the loop that succeeded, otherwise we'll have a memory leak.

This is really a problem, I don't have a deep understanding of the kmalloc's code at the moment, maybe it needs to trouble you to fix it.

vvaltchev: Wow, you did quite some testing to catch that! I'm impressed.

Don't worry about the roll-up, it's not going to be that hard. I'll give you a hint, but I don't have time to write and test it properly in the next few days, so I guess you can do it easily. Here's the loop:

```
   for (int i=(int)h->heap_data_size_log2-1; i >= 0 && tot < desired_size; i--)
   {
      const size_t s = (1u << i);

      if (!(desired_size & s))
         continue;

      addr = internal_kmalloc(h,
                              s,
                              big_block_node,
                              rounded_up_size,
                              true,              /* mark node as allocated */
                              do_actual_alloc);
```

The variable `i` goes from `h->heap_data_size_log2-1` to `0` inclusive. If that call failed, you just need to add another loop in the `if (!addr) {` case like this:

```
if (!addr) {
   for  (++i; i <= (int)h->heap_data_size_log2-1; i++) {
      void *p = big_block + tot;
      size_t s = (1u << i);
      internal_kfree(h,
                     ptr,
                     s /* size */,
                     false /* allow split */,
                     do_actual_alloc);
      tot -= s;
   }
   return NULL;
}
```

That should work. What does it mean? After checking that we have space for the big block, in kmalloc() we alloc smaller and smaller power-of-two sized blocks. If we fail in the internal kmalloc, we need to do exactly the opposite: start from the last block we know we succeeded to alloc, which is `(i+1)`, up to the first one. Each time the size will be still `1 << i`. The pointer is `big_block + tot` where `tot` is the total size allocated so far in small chunks.

I haven't checked that this works of course, but it's 95% that.

aenrbes: Thanks for the hint. But I'll be out of home for the next few days and probably won't be free until Monday . If you're not free by then, I'll get on with it :-)

vvaltchev: Sure, no problem @aenrbes. Thanks for the info!

vvaltchev:
@aenrbes I fixed this bug in https://github.com/vvaltchev/tilck/pull/216, therefore, I'm closing this PR.

PR #218, opened 2024-08-04
Refactoring on the RISCV modules and other things

vvaltchev:
Hello @aenrbes,
just wanted to let you know that I made the new RISCV modules visible in the `sysfs` filesystem.
For that, I also needed to rename them because the module name in Tilck is taken directly from the directory name inside modules and the old names were too generic. 

In other words, while it looked great to have `modules/rtc` and then in `riscv/` all the drivers for RISCV, the problem is that the module directory is one thing for the build system / config and the $ARCH sub-directory is just a way to compile only $ARCH-specific code inside it, not as a convenient way to have a different variation of the same module on different architectures. In other words, most of the modules' code should not be in `modules/<name>/<arch>` but just in `modules/<name>`.  And the $ARCH subdirectory is supposed to be for the special cases, for example when the same HW device have differences across architectures or assembly code etc.

The reason for that design is to allow unit tests to compile and test that noarch code, as they cannot have arch-specific stuff. And, in general, making the C code portable pushes the introduction of the correct abstractions in the whole project.

Said that, I see and fully understand where the confusion comes from: I placed myself multiple drivers in `generic_x86` because I was too lazy to abstract the C code and because I knew that things like kb8042 will never be available for the other architectures anyway. 

In conclusion, this is just a FYI for you to understand the reasoning behind my changes. I'll take care of re-organizing the code around and will probably introduce a special file like `modules/<name>/module_archs` to specify the list of architectures for which the module should be compiled, even most of its C code can be compiled (but not run) everywhere if the HAL is implemented somehow. I'm still thinking about what's best. The goals are:

- as much as _portable_ C code as possible (even if it uses HAL macros of funcs)

- as much as possible C code testable with unit tests, mocks etc. 

- don't compile/link modules and code that does not apply to a given architecture into the kernel, even if it's portable, because it's useless. But compile and link it in the noarch "kernel" for unit tests

Vlad


aenrbes:
Thank you for explaining this in detail.
At the moment all my work has been merged. This was the first time I had contributed to an open source project in such depth, and it helped me a lot in my career. Thank you for your patience and careful review. Next I'll start trying to port tilck to a real-world chip.

vvaltchev:
@aenrbes Thank you for contributing!
Your work has been incredibly valuable to this project and to me. I'm sure that will also boost your career as well because it's a **remarkable contribution** to an operating system project. Very few people could have done it like that, almost completely independently, with little to no guidance.

Any of your work is always welcome here, anytime you want.

Great job!

Cheers,
Vlad

PR #219, opened 2024-08-12
[riscv64] Refactor bsp framework

aenrbes:
Hi @vvaltchev 

I have recently successfully ported tilck to the Sipeed MaixCAM (profile: https://wiki.sipeed.com/hardware/en/maixcam/index.html,  purchase: https://www.aliexpress.com/store/911876460 (It seems that the aliexpress official sipeed store is temporarily out of stock. But I can bought it at taobao official sipeed store)), which is based on the licheerv-nano core board (profile: https://wiki.sipeed.com/hardware/en/lichee/RV_Nano/1_intro.html) and has a 2.3-inch screen and a camera. Of course, you also need a USB to serial module.

tilck-licheerv-nano

About this PR:
Refactor bsp framework, and introduces the BOARD environment variable (default is 'qemu-virt'), which specifies the board level target to be compiled.


vvaltchev:
Hi @aenrbes,
this is an outstanding achievement! Making Tilck to work on real hardware is what really matters at the end of the day! :-)
I'll consider buying the same hardware as well. 

Unfortunately, I'm super tired and cannot make a full review for this change today, but I'll try to do that in the following days.

CMakeLists.txt
```diff
@@ -549,6 +549,18 @@ elseif (${ARCH} STREQUAL "x86_64")
 
 elseif (${ARCH} STREQUAL "riscv64")
 
+   if ("$ENV{BOARD}" STREQUAL "")
```
vvaltchev, line 552: Not a full review, but you shouldn't access the `$BOARD` env variable like that, simply because if it changes on the next build, it will be ignored by CMake. The proper way to that instead is use a "sticky" variable, so that the value of the ENV variable is taken when CMake is run, but if it changes after that, the build will fail. See `define_env_cache_str_var()`. You just need to add an entry for BOARD in this block, after ARCH:
```
# BEGIN [Environment-controlled "sticky" cache variables]

# Main variables
define_env_cache_str_var(TCROOT_PARENT ${CMAKE_SOURCE_DIR})
define_env_cache_str_var(TCROOT ${TCROOT_PARENT}/toolchain2)
define_env_cache_str_var(ARCH ${DEFAULT_TARGET_ARCH})
define_env_cache_bool_var(TEST_GCOV)
define_env_cache_bool_var(KERNEL_GCOV)
````

vvaltchev (changes requested):
Not a full view. I just wanted to share early the comment about the "sticky" variable. 

It's interesting that none of the pre-existing abstractions are not enough so now we need to handle also the concept of BOARD. Your change looks interesting and probably in the right direction, but I'll need some time to review it and think about it, because it's a new thing that has and will have quite the impact on the toolchain/build of the project.

vvaltchev (changes requested):
I made a change, commit 751282bf, that improves the installed status functions of all the toolchain packages, so that we can see the full list architectures for every package, not just the one for the current $ARCH.

Your change looks almost good. So far I think you just need to make BOARD be a sticky env variable as described in the other comment. But I'd need to take a final look after your change is rebased on the top of master.

vvaltchev (approved):
Thanks for rebasing your change and addressing the comments, @aenrbes. The change looks good to me now.

PR #223, opened 2024-08-12
[paging] Using ulong type hw_flags

aenrbes:


vvaltchev:
@aenrbes The change looks good to me, but I don't understand why you need `ulong` for the flags instead of `u32`. Do you need to pass something bigger besides a few bits? Can you explain that in the commit description?

aenrbes:
On the riscv64 architecture, the highest few bits(54-63) are reserved
for vendor defined page attribute bits.
`early_get_cpu_features()` sets these vendor bits at runtime, as shown below

```
   if (f->isa_exts.svpbmt) {

      f->page_mtmask = _PAGE_MTMASK_SVPBMT;
      f->page_cb =     _PAGE_PMA_SVPBMT;
      f->page_wt =     _PAGE_NOCACHE_SVPBMT;
      f->page_io =     _PAGE_IO_SVPBMT;

   } else if (f->vendor_id == THEAD_VENDOR_ID) {

      f->page_mtmask = _PAGE_MTMASK_THEAD;
      f->page_cb =     _PAGE_PMA_THEAD;
      f->page_wt =     _PAGE_NOCACHE_THEAD;
      f->page_io =     _PAGE_IO_THEAD;
   }
```

vvaltchev:
@aenrbes Thank you for the clarification.

PR #225, opened 2024-08-20
[riscv64][bsp] fix broken build_toolchain due to PATH

aenrbes:
@vvaltchev 

I'm sorry that I missed the PATH problem in the build_toolchain. When building licheerv-nano's bootloader, it cannot include the PATH of the toolchain's bin, because it need to use the host system's python instead of the python under the toolchain's bin.

Now I have retested the whole process starting with 'git clone' and everything is working fine.

### DIFF
--- a/other/bsp/riscv64/licheerv-nano/bootloader
+++ b/other/bsp/riscv64/licheerv-nano/bootloader
@@ -5,12 +5,16 @@ function internal_build_bootloader {
 
    echo "Building bootloader... "
    reset_cc_vars_to_null
+   local saved_path="$PATH"
+   export PATH=${PATH//$GCC_TOOLCHAIN:/}
 
    set +e
    run_command2 "source build/cvisetup.sh" build.log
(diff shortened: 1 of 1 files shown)


other/bsp/riscv64/licheerv-nano/bootloader
```diff
@@ -5,12 +5,15 @@ function internal_build_bootloader {
 
    echo "Building bootloader... "
    reset_cc_vars_to_null
+   export PATH=${PATH//$GCC_TOOLCHAIN:/}
```
vvaltchev, line 8: We already have a safer pattern for this case, see `build_libmusl`.
In practice, you need to save the path using a local variable:
```
local saved_path="$PATH"
```
Then you modify it as you want.
At the end of the function, you restore the path with:
```
export PATH="$saved_path"
```
Can you please use that pattern?

vvaltchev (changes requested):

PR #224, opened 2024-08-19
[riscv64][bsp] support sipeed licheerv-nano board

aenrbes:
Hi @vvaltchev,

This PR add support for Sipeed licheerv-nano core board ( _https://wiki.sipeed.com/hardware/en/lichee/RV_Nano/1_intro.html_ )

**BSP usage:**
under tilck project:

1. Make image
```
export ARCH=riscv64
export BOARD=licheerv-nano
./scripts/build_toolchain
make
```
2. Then flash `build/tilck.img` into a microSD card;
3. Insert the microsd card into the card slot on the licheerv-nano board;
4. Connect uart0 on the board (baudrate 115200), and power on the board.
5. Have fun!

boot log:
```
FSBL zIXgg9:g605e5cb6-dirty:2024-08-19T16:32:39+08:00
st_on_reason=d0000
st_off_reason=0
P2S/0x1000/0xc00a200.
SD/0x9200/0x1000/0x1000/0.P2E.
DPS/0xa200/0x2000.
SD/0xa200/0x2000/0x2000/0.DPE.
cv181x DDR init.
ddr_param[0]=0x78075562.
pkg_type=5
D1_3_2
DDR3-2G-QFN
Data rate=1866.
DDR BIST PASS
PLLS.
PLLE.
C2S/0x0/0x0/0x0.
No C906L image.
MS/0xc200/0x80000000/0x1c000.
SD/0xc200/0x1c000/0x1c000/0.ME.
L2/0x28200.
SD/0x28200/0x200/0x200/0.L2/0x414d3342/0xcafe44a4/0x80200000/0x43400/0x43400
COMP/1.
SD/0x28200/0x43400/0x43400/0.DCP/0x80200020/0x1000000/0x81900020/0x43400/1.
DCP/0x9027e/0.
Loader_2nd loaded.
Use internal 32k
Jump to monitor at 0x80000000.
OPENSBI: next_addr=0x80200020 arg1=0x80080000
OpenSBI v0.9
   ____                    _____ ____ _____
  / __ \                  / ____|  _ \_   _|
 | |  | |_ __   ___ _ __ | (___ | |_) || |
 | |  | | '_ \ / _ \ '_ \ \___ \|  _ < | |
 | |__| | |_) |  __/ | | |____) | |_) || |_
  \____/| .__/ \___|_| |_|_____/|____/_____|
        | |
        |_|

Platform Name             : LicheeRv Nano
Platform Features         : mfdeleg
Platform HART Count       : 1
[...]

PR #230, opened 2024-09-18
[riscv64] Fix C function call convention errors in asm_defs.h

aenrbes:
Hi @vvaltchev 

I noticed that the release build of riscv64 crashed in the latest ci test. I did some debug work as follows.

In `asm_save_regs_and_schedule()`, `save_callee_regs()` will save all callee saved regs and then call `do_schedule`, if no sched happened, `do_schedule` will just return without execute `resume_callee_regs()`. In this case, the value of the `s1` register will change (but according to the C function call convention, the `s1` register must be restored before returning to the caller).

This is why selftest crashed in the case of release build. When `asm_save_regs_and_schedule()` returns to the caller (`save_regs_and_schedule()`), the value of `s1` has changed. But the caller (that is, `save_regs_and_schedule()`) still assumes that `s1` has not changed at all, thus creating a series of exceptions.

### DIFF
--- a/include/tilck/kernel/arch/riscv/asm_defs.h
+++ b/include/tilck/kernel/arch/riscv/asm_defs.h
@@ -85,8 +85,8 @@
    REG_S s10, 26 * RISCV_SZPTR(sp)
    REG_S s11, 27 * RISCV_SZPTR(sp)
 
-   csrr s1, sstatus
-   REG_S s1, 33 * RISCV_SZPTR(sp)
+   csrr t0, sstatus
+   REG_S t0, 33 * RISCV_SZPTR(sp)
 
 .endm
 
@@ -150,20 +150,20 @@
    REG_S x30, 30 * RISCV_SZPTR(sp)
    REG_S x31, 31 * RISCV_SZPTR(sp)
 
-   csrr s0, sepc
-   csrr s1, sstatus
-   csrr s2, stval
-   csrr s3, scause
-   csrr s4, sscratch
+   csrr t0, sepc
+   csrr t1, sstatus
+   csrr t2, stval
+   csrr t3, scause
+   csrr t4, sscratch
 
-   REG_S s0, 32 * RISCV_SZPTR(sp)
-   REG_S s1, 33 * RISCV_SZPTR(sp)
-   REG_S s2, 34 * RISCV_SZPTR(sp)
-   REG_S s3, 35 * RISCV_SZPTR(sp)
-   REG_S s4, 38 * RISCV_SZPTR(sp)
+   REG_S t0, 32 * RISCV_SZPTR(sp)
+   REG_S t1, 33 * RISCV_SZPTR(sp)
+   REG_S t2, 34 * RISCV_SZPTR(sp)
+   REG_S t3, 35 * RISCV_SZPTR(sp)
+   REG_S t4, 38 * RISCV_SZPTR(sp)
 
-   addi s0, sp, SIZEOF_REGS
-   REG_S s0,  2 * RISCV_SZPTR(sp)
+   addi t0, sp, SIZEOF_REGS
+   REG_S t0,  2 * RISCV_SZPTR(sp)
 
 .endm
 


aenrbes:
Now it looks like all riscv64 CI tests is passed. By the way, I would like to remind you that I have uploaded some videos and photos of riscv64 running on sipeed maixcam board in #224

aenrbes:
I dug a little deeper into the issue and found that the same problem existed in the `panic_save_current_state()`->`save_all_regs()` call, which was the cause of `fault in panic`.

aenrbes:
I dug a little deeper into the issue and found that the same problem existed in the `panic_save_current_state()`->`save_all_regs()` call, which was the cause of `fault in panic`.

vvaltchev:
> Hi @vvaltchev
> 
> I noticed that the release build of riscv64 crashed in the latest ci test. I did some debug work as follows.
> 
> In `asm_save_regs_and_schedule()`, `save_callee_regs()` will save all callee saved regs and then call `do_schedule`, if no sched happened, `do_schedule` will just return without execute `resume_callee_regs()`. In this case, the value of the `s1` register will change (but according to the C function call convention, the `s1` register must be restored before returning to the caller).
> 
> This is why selftest crashed in the case of release build. When `asm_save_regs_and_schedule()` returns to the caller (`save_regs_and_schedule()`), the value of `s1` has changed. But the caller (that is, `save_regs_and_schedule()`) still assumes that `s1` has not changed at all, thus creating a series of exceptions.

Thank you so much @aenrbes, for debugging and fixing the issue!

vvaltchev:
> Now it looks like all riscv64 CI tests is passed. By the way, I would like to remind you that I have uploaded some videos and photos of Tilck on riscv64 running on Sipeed MaixCAM board in #224

@aenrbes I haven't forgotten about that, at all. That's why I spent some time across several weekends to make the CI tests to work for RISCV64 as well. We're approaching the stability needed for RISCV64 to be considered as "supported". Let me check if multiple CI runs will be stable or some other subtle bug will come out.

Other than that, did the framebuffer console worked as-it-is on Sipeed MaixCAM ? I remember you mentioned that on QEMU it couldn't work because of the color mode required and that somehow wasn't supported by the current implementation. I'd like users to be able to build & run Tilck RISCV64 and observe the same thing they do on the screenshots. But to get there the FB console needs to work on QEMU on the new architecture too. Can you at least re-share with me the code snippet to make the FB console work on QEMU too? 

Side note: I care about Tilck moving forward and advertise the new architecture but, as mentioned previously, the time I can spend on the project is very limited, so that's why it's taking a long time for me to do anything, in particular when subtle bugs come out. _[ In this case, after I switched to new testing mode where the test script writes to the serial port literally to shell's command line, I found out that 20-20% of the jobs failed due to a timeout because the shell got no input and just hanged. That new test mode was unusable. After quite some debugging it turned out that there was a race condition in select.c and poll.c, plus other smaller bugs. I had to fix all of those, in order to make the tests run like that, even just on i386. ]_

I cannot just rush the things up and make big claims until I'm confident that the quality level is good enough and on par with the rest of the project. If at any time, for any reason, you'd like (or need) things to move faster, let's have a chat and figure out a way for you to step in and help to make that happen. Without external help, I cannot make any commitments for the project and timelines.

We can probably sync better via email, not just using Github issues and pull requests, to talk about this things and plan. You can find my email in the commit messages.

Thanks,
Vlad

aenrbes:
Hi @vvaltchev 
First of all, framebuffer mode does not work directly on MaixCam and requires a patch work on the vendor released bootloader (which I provided in #224 ). And need to add a 24bit bmp file named 'logo.bmp' to the bootpart of the image. But I'm not sure if it's necessary to integrate patch work into the bootloader build script.

In addition to qemu, I have never enabled the framebuffer mode, because u-boot does not have any support for qemu framebuffer devices (mainly ramfb or virio-gpu-device). Of course, I could try porting a qemu framebuffer driver to the u-boot if necessary.

aenrbes:
@vvaltchev 

I have successfully ported the ramfb driver to the u-boot and enabled tilck's fb mode, but unfortunately qemu's ramfb doesn't have any hardware acceleration so it doesn't work smoothly (at a typical 800*600 resolution).
The work required to enable Tilck's fb console: 
1. Patch u-boot; 
2. Remove the `-sercon` option from uEnv.txt bootargs. 
3. In the run_qemu script, remove the `-nographic` option and add the `-serial stdio -vga none-device ramfb` option

patch file:
```
--- toolchain2/riscv64/u-boot/arch/riscv/lib/bootm.c	2024-04-02 18:58:58.000000000 +0800
+++ toolchain2/riscv64/u-boot-new/arch/riscv/lib/bootm.c	2024-09-19 23:27:04.304732201 +0800
@@ -22,9 +22,119 @@
 #include <dm/device.h>
 #include <dm/root.h>
 #include <u-boot/zlib.h>
+#include <asm/io.h>
 
 DECLARE_GLOBAL_DATA_PTR;
 
+#define BE16(x) __builtin_bswap16(x)
+#define BE32(x) __builtin_bswap32(x)
+#define BE64(x) __builtin_bswap64(x)
+
+#define QEMU_FRAMEBUFFER_ADDR (0x80000000UL + 0x6400000UL)
+#define QEMU_FRAMEBUFFER_WIDTH 400
+#define QEMU_FRAMEBUFFER_HEIGHT 300
+
+#define QEMU_FWCFG_BASE 0x10100000UL
+#define QEMU_FWCFG_SELECTOR 0x10100008UL
+#define QEMU_FWCFG_DATA 0x10100000UL
+#define QEMU_FWCFG_ADDR 0x10100010UL
+
+#define fourcc_code(a, b, c, d) ((uint32_t)(a) | ((uint32_t)(b) << 8) | \
+            ((uint32_t)(c) << 16) | ((uint32_t)(d) << 24))
+
+#define QEMU_RAMFB_RGB888   fourcc_code('R', 'G', '2', '4')
+#define QEMU_RAMFB_XRGB8888 fourcc_code('X', 'R', '2', '4')
+#define QEMU_RAMFB_RGB565   fourcc_code('R', 'G', '1', '6')
+
+struct qemu_ramfb_cfg {
+   uint64_t addr;
+   uint32_t format;
+   uint32_t flags;
+   uint32_t width;
+   uint32_t height;
+   uint32_t stride;
+} __attribute__((packed));
+
+struct qemu_fwcfg_dma_access {
+   uint32_t cntl;
+   uint32_t len;
+   uint64_t addr;
+} __attribute__((packed));
+
+struct qemu_fwcfg_file {
+   uint32_t size;
+   uint16_t select;
+   uint16_t reserved;
+   char name[56];
+} __attribute__((packed));
+
+static void qemu_fwcfg_dma_trans(void *addr, uint32_t len, uint32_t cntl) {
+   struct qemu_fwcfg_dma_access access;
+
+   access.cntl = BE32(cntl);
+   access.len = BE32(len);
+   access.addr = BE64((uint64_t)addr);
+
+   writeq((BE64((uint64_t)&access)), (void *)QEMU_FWCFG_ADDR);
+
+   while (BE32(access.cntl) & (uint32_t)(~0x1)) {
+   }
+}
+
+static void qemu_fwcfg_dma_sel_read(void *buf, uint32_t sel, uint32_t len) {
+   qemu_fwcfg_dma_trans(buf, len, (sel << 16) | 0x08 | 0x02);
+}
+
+static void qemu_fwcfg_dma_sel_write(void *buf, uint32_t sel, uint32_t len) {
+   qemu_fwcfg_dma_trans(buf, len, (sel << 16) | 0x08 | 0x10);
+}
+
+static void qemu_fwcfg_dma_read(void *buf, int len) {
+   qemu_fwcfg_dma_trans(buf, len, 0x02);
+}
+
+static int qemu_ramfb_setup(void *fdt)
+{
+   int offset;
+   uint32_t count = 0, select = 0;
+   struct qemu_fwcfg_file file;
+
+   qemu_fwcfg_dma_sel_read(&count, 0x19, sizeof(count));
+   count = BE32(count);
+
+   for (int i = 0; i < count; i++) {
+      qemu_fwcfg_dma_read(&file, sizeof(file));
+      if (!strcmp(file.name, "etc/ramfb")) {
+         select = BE16(file.select);
+         break;
+      }
+   }
+
+   struct qemu_ramfb_cfg ramfb_cfg = {
+      .addr = BE64(QEMU_FRAMEBUFFER_ADDR),
+      .format = BE32(QEMU_RAMFB_RGB888),
+      .flags = BE32(0),
+      .width = BE32(QEMU_FRAMEBUFFER_WIDTH),
+      .height = BE32(QEMU_FRAMEBUFFER_HEIGHT),
+      .stride = BE32(3 * QEMU_FRAMEBUFFER_WIDTH)
+   };
+
+   qemu_fwcfg_dma_sel_write(&ramfb_cfg, select, sizeof(ramfb_cfg));
+
+   for (int i = 0; i < 3 * QEMU_FRAMEBUFFER_WIDTH * QEMU_FRAMEBUFFER_HEIGHT;) {
+      *(char *)(QEMU_FRAMEBUFFER_ADDR + i) = 0xff; i++;
+      *(char *)(QEMU_FRAMEBUFFER_ADDR + i) = 0; i++;
+      *(char *)(QEMU_FRAMEBUFFER_ADDR + i) = 0; i++;
+   }
+
+   offset = fdt_add_subnode(fdt, 0, "simplefb");
+   if (offset < 0)
+      return -1;
+
+   fdt_setprop_string(fdt, offset, "compatible", "simple-framebuffer");
+   return fdt_setup_simplefb_node(fdt, offset, QEMU_FRAMEBUFFER_ADDR, QEMU_FRAMEBUFFER_WIDTH, QEMU_FRAMEBUFFER_HEIGHT, 3 * QEMU_FRAMEBUFFER_WIDTH, "r8g8b8");
+}
+
 __weak void board_quiesce_devices(void)
 {
 }
@@ -91,6 +201,8 @@
 	debug("## Transferring control to kernel (at address %08lx) ...\n",
 	      (ulong)kernel);
 
+   qemu_ramfb_setup((void *)images->ft_addr);
+
 	announce_and_cleanup(fake);
 
 	if (!fake) {

```

qemu-fb
test
qemu-riscv-fb

vvaltchev:
That's amazing, @aenrbes. Thank you for the patch!
As soon as I have some time, I'll try that patch out myself.

I think that the framebuffer is slow because we're running on pure software emulation? Also, I see that the FB memory it's not accessed through write-combining memory (WC). On real HW, in particular modern x86_64 machines (Dell XPS 13" from 2017), it made a huge difference when I discovered the WC mode and used PAT to achieve that. Literally 200x difference when writing 32-bit words. That's not a random number, indeed. Previously each 32-bit write, had to go directly to the memory, which costs 200+ cycles. 

I noticed that on RISCV64 we don't have WC:
```
void set_pages_pat_wc(pdir_t *pdir, void *vaddr, size_t size)
{
   ASSERT(IS_L0_PAGE_ALIGNED(vaddr));
   ASSERT(IS_L0_PAGE_ALIGNED(size));

   const void *end = vaddr + size;

   while (vaddr < end) {

      if (in_big_page(pdir, vaddr)) {
         set_big_page_attr(pdir, vaddr, _PAGE_WT);
         vaddr += L1_PAGE_SIZE;
         continue;
      }

      set_4kb_page_attr(pdir, vaddr, _PAGE_WT);
      vaddr += PAGE_SIZE;
   }
}
```
And you used WT instead. Do you know if WC mappings even exist on RISCV ? With WC memory we have no consistency guarantees and the CPU is allowed to re-order in any way the stores. That's perfect for framebuffers. 

How fast was the FB on the real HW you tried compared to QEMU ?

aenrbes:
@vvaltchev 

About first question, I've noticed that several of the ramfb presentations mention that qemu's ramfb itself is inefficient, and not recommended for use in formal products.

About WC mapping, riscv itself does not directly define the WC. Riscv page property 's define is as following code snippet (from Linux/arch/riscv/include/asm/pgtable-64.h)

```
/*
 * [63:59] T-Head Memory Type definitions:
 * bit[63] SO - Strong Order
 * bit[62] C - Cacheable
 * bit[61] B - Bufferable
 * bit[60] SH - Shareable
 * bit[59] Sec - Trustable
 * 00110 - NC   Weakly-ordered, Non-cacheable, Bufferable, Shareable, Non-trustable
 * 01110 - PMA  Weakly-ordered, Cacheable, Bufferable, Shareable, Non-trustable
 * 10010 - IO   Strongly-ordered, Non-cacheable, Non-bufferable, Shareable, Non-trustable
 */
#define _PAGE_PMA_THEAD	((1UL << 62) | (1UL << 61) | (1UL << 60))
#define _PAGE_NOCACHE_THEAD	((1UL << 61) | (1UL << 60))
#define _PAGE_IO_THEAD		((1UL << 63) | (1UL << 60))
#define _PAGE_MTMASK_THEAD	(_PAGE_PMA_THEAD | _PAGE_IO_THEAD) | (1UL << 59))
```

As you can see, as long as SO(Strong Order) and C(Cacheable) are not defined, then the same effect as WC is achieved, as following linux code(from Linux/arch/riscv/include/asm/pgtable.h)

```
#define pgprot_writecombine pgprot_writecombine
static inline pgprot_t pgprot_writecombine(pgprot_t _prot)
{
	unsigned long prot = pgprot_val(_prot);

	prot &= ~_PAGE_MTMASK;
	prot |= _PAGE_NOCACHE;

	return __pgprot(prot);
}
```

About FB on the real HW, as in the #224  video, I didn't feel any noticeable lack of fluency. And I didn't feel any stutters on qemu with 400*300 resolution.

vvaltchev:
@aenrbes Thank you for the info! I'm glad that on real HW there is no problem and that we're using the proper memory type. It's OK if it's slow on QEMU as long as it's not our fault :-)

P.S. I had to edit your comment and remove a code snippet from a project with an incompatible license. Please don't quote code from other projects. I don't want Tilck to have some day copyright issues.

vvaltchev:
@aenrbes I just noticed that your patch for the framebuffer support requires touching u-boot's source code, which is an external component. Why is that? Couldn't Tilck detect QEMU's ramfb device by scanning the device tree? I'm absolutely against forking and patching u-boot. But at the same time, we cannot avoid the FB support.

Also, even in U-BOOT, why is QEMU_FRAMEBUFFER_ADDR a hard-coded constant?

I spent some time thinking about claiming an official support for RISCV64 and I'm not comfortable with that yet, despite quite the additional effort on my side in the last months. I'm not happy with a half-baked support for RISCV64 and in particular I'm not happy with the screenshots showing the framebuffer support on both QEMU and the HW device, while that requires touching U-BOOT and users would have no easy way to try that. Also, I'm not happy with the zero support for the pre-existing applications. At least some of them should work on RISCV64 as well. For example, `lua` and `micropython`. Ideally, all of them should work, but I realize that's too much and some apps like TCC and VIM are not so critical. TODO work:

- We need FB support in the kernel at least for QEMU, implemented in a generic way, with automatic detection for the QEMU ramfb device somehow, probably using RISCV's device tree. No U-BOOT hacks that would require a fork plus a lot of effort to support, no hard-coded memory addresses for the framebuffer.

- We need at least LUA and MICROPYTHON to work. LUA builds, but somehow does not read from stdin any keyboard input, with the current serial console. Micropython does not even build on RISCV64. I know that's a nasty business to get 3rd party apps build with a cross-compiler, but it's very useful for people using the new architecture.

- We need some documentation written for RISCV64. For example, we'd need a description of all the HW components supported (virtual and non-virtual, like the `goldfish` RTC, the multiple IRQ controllers supported etc. Also, we need a description of the boards supported. As an example of what's needed, check this: https://github.com/vvaltchev/tilck?tab=readme-ov-file#hardware-support. This will be kind of hard for me to write since I don't have knowledge on the RISCV world in order to put the current Tilck support in the right perspective. 

- Optionally, we need to clear at least some of the tech debt. For example, I managed to share most of the `switch_to_task()` code between the two architectures, but I still needed to keep separate `arch_usermode_task_switch()` because on RISCV64 `restore_fpu_regs()` contains already the instruction to enable the FPU, while on x86 we do it in a dedicated step in the C code. Can we have the RISCV64 code do the same? There are more examples of code that really should be shared.

@aenrbes Are you willing to help with any of those tasks, so that we get "official RISCV64 support" sooner?

Thank you,
Vlad

PR #231, opened 2024-11-18
[riscv64] Add support for qemu-ramfb and enable framebuffer

aenrbes:
Hi @vvaltchev 

According to the suggestion you mentioned in #230, I ported the driver of qemu-ramfb to tilck kernel and enabled the framebuffer mode by default.


aenrbes:
Hi @vvaltchev 

My preliminary description of the hardware support for riscv is as follows:

riscv64 support

Tilck's riscv64 architecture mainly focuses on embedded applications. Tilck runs in RV64 supervisor mode (Smode), uses 
SV39 three-level page table virtual memory (the code architecture can theoretically support any multilevel page table), 
and supports user space floating point.

hardware peripheral support

riscv64 architecture adopts a device tree-based driver model, and all hardware information will be obtained from the 
device tree transmitted by the bootloader. Therefore, tilck riscv kernel code does not have any board-level description 
code. The system tick clock invokes the SBI monitor program (openSBI) through riscv SBI interface Instead of 
implementing a driver in tilck; The system RTC clock implemented a very simple goldfish virtual RTC driver, of course, 
goldfish rtc can only be used on the qemu virtual platform; Tilck riscv does not implement PC peripherals such as 
keyboards, the input and output all rely on serial ports (which is enough for most embedded applications), the most 
common ns16550 driver is implemented; To better manage the multi-level nested interrupt architecture common in 
embedded chips, tilck's riscv implements a simplified framework similar to the idea of Linux irq domain, the most 
common intc and plic interrupt controller drivers in riscv are realized.

kernel's cmdline

The modification of kernel's cmdline for riscv64 is a little more complicated than that for i386, requiring the 
modification of the bootargs parameter in uEnv.txt(located in the other/bsp/corresponding board subfolder).

board support

riscv64 architecture implements a very simple bsp framework, which only needs to set the environment variable 
`BOARD` to build a image of the corresponding board. Currently only support sipeed licheerv-nano board (https://wiki.sipeed.com/hardware/en/lichee/RV_Nano/1_intro.html) and qemu-virt virtual board (the default).
qemu-virt has the framebuffer mode enabled by default, while licheerv-nano needs to patch original bootloader 
provided by sipeed to use the framebuffer.

The use of qemu-virt board is very simple, just `export ARCH=riscv64`, and follow instructions of i386's.

Using the licheerv-nano is a little more cumbersome:

1. Make image
```
export ARCH=riscv64
export BOARD=licheerv-nano
./scripts/build_toolchain
make
```
2. Flash the image file build/tilck.img into a microSD card;
3. Insert the microsd card into the card slot on the licheerv-nano board;
4. Connect uart0 on the board (baudrate 115200), and power on the board.
5. Have fun!

vvaltchev:
> Hi @vvaltchev
> 
> According to the suggestion you mentioned in #230, I ported the driver of qemu-ramfb to tilck kernel and enabled the framebuffer mode by default.

Thanks for working on this @aenrbes. It looks very cool.
However, I tried this change with QEMU and noticed two things:

1. no keyboard input on the FB window (key strokes have no effect)
2. the FB resolution is super low. Can we increase this somehow?

I guess that fixing issue 1 requires also a driver for whatever KB device QEMU supports. Is that correct? Having support for qemu-ramfb without keyboard input mostly defeats the purpose of the FB, even if it's still a step forward. Now we don't need to hack U-BOOT, thanks to your change. How hard would be to have KB input working?

For issue 2, is there any easy way to increase the resolution?

Thank you,
Vlad

modules/ramfb/riscv/ramfb.c
```diff
@@ -0,0 +1,196 @@
...
+{
+   struct qemu_fwcfg_dma_access access;
+
+   access.cntl = BE32(cntl);
+   access.len = BE32(len);
+   access.addr = BE64((u64)addr);
+   mmio_writeq((BE64((u64)&access)), (void *)QEMU_FWCFG_ADDR);
+
+   while (BE32(access.cntl) & (u32)(~0x1))
+      ;
```
vvaltchev, line 69: Style: in Tilck the preferred style for that is:
```
while (...) { }
```

vvaltchev:
I'm adding the warning message myself: https://github.com/vvaltchev/tilck/pull/235

aenrbes:
> I guess that fixing issue 1 requires also a driver for whatever KB device QEMU supports. Is that correct? Having support for qemu-ramfb without keyboard input mostly defeats the purpose of the FB, even if it's still a step forward. Now we don't need to hack U-BOOT, thanks to your change. How hard would be to have KB input working?

Getting KB input to work on riscv will be a huge job for now. Because riscv qemu-virt platform does not have a PS/2 controller, there is no way to simply adapt a keyboard driver to make it work.

> For issue 2, is there any easy way to increase the resolution?

Modify the `setupfb` parameter in `uEnv.txt` in the bsp board subdirectory

`setupfb=setexpr simplefb_addr @KERNEL_LOAD@ + 0x900000; fdt mknode / simplefb; fdt set /simplefb compatible "simple-framebuffer"; fdt set /simplefb width <0x190>; fdt set /simplefb height <0x12c>; fdt set /simplefb stride <0x4b0>; fdt set /simplefb format "r8g8b8"; fdt set /simplefb reg <0x0 0x${simplefb_addr} 0x0 0x0>`

Change the value of `fdt set /simplefb width <0x190>; fdt set /simplefb height <0x12c>; fdt set /simplefb stride <0x4b0>`，where `stride` is equal to `width*3`

vvaltchev:
> Getting KB input to work on riscv will be a huge job for now. Because riscv qemu-virt platform does not have a PS/2 controller, there is no way to simply adapt a keyboard driver to make it work.

Thank you for the comments, makes sense.


> Change the value of `fdt set /simplefb width <0x190>; fdt set /simplefb height <0x12c>; fdt set /simplefb stride <0x4b0>`，where `stride` is equal to `width*3`

Thanks for that!

PR #237, opened 2024-11-22
Update README.md with info about RISCV64 support

vvaltchev:
@aenrbes This PR adds some basic information about the RISCV64 support in Tilck in the main README file. A bit more stuff will follow like the instructions to build for the licheerv-nano board, screenshots etc. RISCV64 will have a dedicated wiki page and README will have a link towards that.

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,13 +29,13 @@ Contents
   - [What Tilck is NOT ?](#what-tilck-is-not-)
     * [Tilck vs Linux](#tilck-vs-linux)
 * [Features](#features)
-   - [Hardware support](#hardware-support)
+   - [i686 support](#i686-support)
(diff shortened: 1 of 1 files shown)


aenrbes:
@vvaltchev I am very pleased that the porting of riscv64 has reached this stage. If there's anything you need, I'm here to help. I am currently considering whether I can write a board-level porting guide for riscv64, so that embedded developers can easily port tilck to other boards without deep understanding of the inner workings of tilck.

vvaltchev:
@aenrbes I'm very pleased as well! Your contribution made it happen.
I like the idea of the guide. The `docs/` directory is the right place for that. We can put a link towards it from both the main README file and from the dedicated wiki page for RISCV (that I'm yet to create).

PR #238, opened 2025-09-10
[null] Implement /dev/[full,null,zero]

vkoskiv:
Hello. Cool project, first of all! I've really enjoyed browsing the source. Bonus points for a solid build system.

You may consider this PR a suggestion - I enjoyed implementing it, but if it doesn't fit the project goals/directions for whatever reason, don't hesitate to close this PR.

Few clarifications/notes to check before merging:

- [ ] I implemented this under `modules/`, following precedent set by `modules/fb`. I can move this under `kernel/` if you think that makes more sense.
- [ ] I'm not sure what the intended way to add tests for module code is, so I just added the tests under `tests/system` and I'm passing `MOD_null` to the `enabled` field in `CMD_ENTRY` so the tests are automatically disabled if `MOD_null == OFF`.
- [ ] Adding these new entries under `/dev` broke the `fs6` test in `tests/system/test_fs1.c`. I fixed that for now by bumping `n_files` from 6 -> 13, but the test code here may be in need of some fixes. As far as I can tell, `dposs` in `generic_fs_dir_seek_test()` was getting overrun even before my changes, with 10 entries under `/dev` >= `n_files` (6). Running the test on my host system under `valgrind(1)` confirmed this overrun.


tests/system/test_mem.c
```diff
@@ -0,0 +1,85 @@
+#include <stdio.h>
+#include <string.h>
+#include <unistd.h>
+#include <fcntl.h>
+
+#include "devshell.h"
+
+void cmd_dev_null(int argc, char **argv) {
```
vvaltchev, line 8: Can you follow the 3-space indentation for this code, so that's aligned with the rest of the project?

CMakeLists.txt
```diff
@@ -397,6 +397,7 @@ list(
    console
    tracing
    sysfs
+   mem
```
vvaltchev, line 400: `mem` doesn't look like the right name for this module, as it doesn't expose `/dev/mem`. Can you rename it to `null` or `zero` ?

vvaltchev (changes requested):
The change looks good to me overall, except for the module name itself, see the comment.

**Note**: I've fixed the overrun in the test function, please rebase on the top of tree.

modules/mem/mem.c
```diff
@@ -0,0 +1,112 @@
...
+static const struct file_ops static_ops_full = {
+   .read = zero_read,
+   .write = full_write,
+   .seek = null_seek,
+};
+
+static int
+mem_create_device_file(int minor,
+                        enum vfs_entry_type *type,
+                        struct devfs_file_info *nfo)
```
vvaltchev, line 67: The params should be aligned with 1st param (`int minor`).

tests/system/test_mem.c
vvaltchev: Thanks!

CMakeLists.txt
vkoskiv: Thanks, fixed!

